Issue #7, opened 2015-06-25
Pester test not passing


BladeFireLight:
Errors from both xFirewall and xIPAddress tests
Describing Schema Validation for MSFT_xFirewall
 [-] Error occurred in Describe block 50ms
   Access to the path 'MSFT_xADRecycleBin' is denied.
   At H:\GitHub\xNetworking\Tests\MSFT_xFirewall.Tests.ps1:10 char:5

Import-Module : The specified module 'xDSCResourceDesigner' was not loaded because no valid module file was found in any 
module directory.
At H:\GitHub\xNetworking\Tests\MSFT_xIPAddress.Tests.ps1:3 char:5
+     Import-Module -Name xDSCResourceDesigner
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ResourceUnavailable: (xDSCResourceDesigner:String) [Import-Module], FileNotFoundException
    + FullyQualifiedErrorId : Modules_ModuleNotFound,Microsoft.PowerShell.Commands.ImportModuleCommand

BladeFireLight:
Describing Schema Validation for MSFT_xFirewall
[10,5: Copy-Item] Access to the path 'MSFT_xADRecycleBin' is denied.
[10,5: Copy-Item] Access to the path 'C:\Program Files\WindowsPowerShell\Modules\MSFT_xADRecycleBin' is denied.
[10,5: Copy-Item] Access to the path 'C:\Program Files\WindowsPowerShell\Modules\MSFT_xADRecycleBin' is denied.
[10,5: Copy-Item] Access to the path 'C:\Program Files\WindowsPowerShell\Modules\MSFT_xADRecycleBin' is denied.
[10,5: Copy-Item] Access to the path 'C:\Program Files\WindowsPowerShell\Modules\MSFT_xADRecycleBin' is denied.
[10,5: Copy-Item] Access to the path 'MSFT_xADRecycleBin' is denied.
[10,5: Copy-Item] Access to the path 'C:\Program Files\WindowsPowerShell\Modules\MSFT_xADRecycleBin' is denied.
[10,5: Copy-Item] Access to the path 'C:\Program Files\WindowsPowerShell\Modules\MSFT_xADRecycleBin' is denied.
[10,5: Copy-Item] Access to the path 'MSFT_xADRecycleBin' is denied.
[...]

PR #10, opened 2015-08-03
Modifying pester tests to not copy file to Program Files

tysonjhayes:
Partial fix for #7.

This will not fix the fact that the user will need xDSCResourceDesigner to run the tests.


tysonjhayes:
I'm unclear why this failed its tests if anyone has any ideas I'd appreciate hearing them.

PR #11, opened 2015-08-07
Test all parameters

### LINKED ISSUE #9: Should xIPAddress disable DHCP?
If a user is setting the IP address via DSC, then it seems reasonable that they don't want DHCP enabled. Currently the test for xIPAddress only checks whether the IP is correct, not whether the interface got that IP via DHCP. I ran into this when trying to convert an interface from DHCP to a static IP, but I wanted to use the same IP that had been assigned to it via DHCP. Right now, it seems I can't use xIPAddress to accomplish this?

lorengordon:
Previously, xIPAddress only checked whether the configured IP Address matched the IPAddress parameter. This pull request checks all parameters against the existing configuration. 

Fixes #9

### DIFF
--- a/DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1
+++ b/DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1
@@ -3,8 +3,6 @@
  #  Address, by accepting values among those given in MSDSCPack_IPAddress.schema.mof
  #######################################################################################>
 
-
-
 ######################################################################################
 # The Get-TargetResource cmdlet.
 # This function will get the present list of IP Address DSC Resource schema variables on the system
@@ -31,9 +29,9 @@ function Get-TargetResource
         [String]$AddressFamily = "IPv4"
     )
 
-
     $returnValue = @{
-        IPAddress = [System.String]::Join(", ",(Get-NetIPAddress -InterfaceAlias $InterfaceAlias -AddressFamily $AddressFamily).IPAddress)
+        IPAddress = [System.String]::Join(", ",(Get-NetIPAddress -InterfaceAlias $InterfaceAlias `
+            -AddressFamily $AddressFamily).IPAddress)
         SubnetMask = $SubnetMask
         DefaultGateway = $DefaultGateway
         AddressFamily = $AddressFamily
@@ -69,7 +67,6 @@ function Set-TargetResource
         [String]$AddressFamily = "IPv4"
     )
 
-
     ValidateProperties @PSBoundParameters -Apply
 }
 
@@ -102,7 +99,6 @@ function Test-TargetResource
     ValidateProperties @PSBoundParameters
 }
 
-
 #######################################################################################
 #  Helper function that validates the IP Address properties. If the switch parameter
 # "Apply" is set, then it will set the properties after a test
@@ -134,14 +130,19 @@ function ValidateProperties
 
     if(!([System.Net.Ipaddress]::TryParse($ip, [ref]0)))
     {
-       throw "IP Address *$IPAddress* is not in the correct format. Please correct the ipaddress in the configuration and try again"
+        throw ( @(
+            "IP Address *$IPAddress* is not in the correct format. Please correct the ipaddress "
+            "in the configuration and try again"
+            ) -join "" | Out-String
+        )
     }
 
     try
     {
         Write-Verbose -Message "Checking the IPAddress ..."
         #Get the current IP Address based on the parameters given.
-        $currentIP = Get-NetIPAddress -InterfaceAlias $InterfaceAlias -AddressFamily $AddressFamily -ErrorAction Stop
+        $currentIP = Get-NetIPAddress -InterfaceAlias $InterfaceAlias -AddressFamily `
+            $AddressFamily -ErrorAction Stop
 
         # Build hash table of current settings
         $currentSettings = @{}
@@ -155,7 +156,11 @@ function ValidateProperties
         #Test if the IP Address passed is equal to the current ip address
         if(!$currentSettings['IPAddress'].Equals($IPAddress))
         {
-            Write-Verbose -Message "IPAddress does NOT match desired state. Expected $IPAddress, actual $($currentSettings['IPAddress'])."
+            Write-Verbose -Message ( @(
+                "IPAddress does NOT match desired state. Expected $IPAddress, actual "
+                "$($currentSettings['IPAddress'])."
+                ) -join "" | Out-String
+            )
             $requiresChanges = $true
         }
         else
@@ -166,7 +171,11 @@ function ValidateProperties
         #Test if the Subnet Mask passed is equal to the current subnet mask
         if(!$currentSettings['PrefixLength'].Equals([byte]$SubnetMask))
         {
-            Write-Verbose -Message "Subnet mask does NOT match desired state. Expected $SubnetMask, actual $($currentSettings['PrefixLength'])."
+            Write-Verbose -Message ( @(
+                "Subnet mask does NOT match desired state. Expected $SubnetMask, actual "
+                "$($currentSettings['PrefixLength'])."
+                ) -join "" | Out-String
+            )
             $requiresChanges = $true
         }
         else
@@ -180,15 +189,25 @@ function ValidateProperties
             $gateway = $DefaultGateway
             if(!([System.Net.Ipaddress]::TryParse($gateway, [ref]0)))
             {
-               throw "Default Gateway *$DefaultGateway* is NOT in the correct format. Please correct the gateway ipaddress in the configuration and try again"
+                throw ( @(
+                    "Default Gateway *$DefaultGateway* is NOT in the correct format. Please "
+                    "correct the gateway ipaddress in the configuration and try again"
+                    ) -join "" | Out-String
+                )
             }
 
             $netIpConf = Get-NetIPConfiguration -InterfaceAlias $InterfaceAlias -ErrorAction Stop
-            $currentSettings['DefaultGateway'] = ($netIpConf."$($AddressFamily)DefaultGateway").NextHop
+            $currentSettings['DefaultGateway'] = (
+                $netIpConf."$($AddressFamily)DefaultGateway"
+                ).NextHop
 
             if(!$currentSettings['DefaultGateway'].Equals($DefaultGateway))
             {
-                Write-Verbose -Message "Default gateway does NOT match desired state. Expected $DefaultGateway, actual $($currentSettings['DefaultGateway'])."
+                Write-Verbose -Message ( @(
+                    "Default gateway does NOT match desired state. Expected $DefaultGateway, "
+                    "actual $($currentSettings['DefaultGateway'])."
+                    ) -join "" | Out-String
+                )
                 $requiresChanges = $true
             }
             else
@@ -198,7 +217,8 @@ function ValidateProperties
         }
 
         #Test if DHCP is already disabled
-        if(!(Get-NetIPInterface -InterfaceAlias $InterfaceAlias -AddressFamily $AddressFamily).Dhcp.ToString().Equals('Disabled'))
+        if(!(Get-NetIPInterface -InterfaceAlias $InterfaceAlias -AddressFamily `
+            $AddressFamily).Dhcp.ToString().Equals('Disabled'))
         {
             Write-Verbose -Message "DHCP is NOT disabled."
             $requiresChanges = $true
@@ -210,20 +230,31 @@ function ValidateProperties
 
         if($requiresChanges)
         {
-            #Apply is true in the case of set - target resource - in which case, it will apply the required IP configuration
+            # Apply is true in the case of set - target resource - in which case, it will apply the
+            # required IP configuration
             if($Apply)
             {
-                Write-Verbose -Message "At least one setting differs from the passed parameters. Applying configuration..."
+                Write-Verbose -Message ( @(
+                    "At least one setting differs from the passed parameters. Applying "
+                    "configuration..."
+                    ) -join "" | Out-String
+                )
 
                 # Build parameter hash table
                 $Parameters = @{}
                 $Parameters["IPAddress"] = $IPAddress
                 $Parameters["PrefixLength"] = $SubnetMask
                 $Parameters["InterfaceAlias"] = $currentIP[0].InterfaceAlias
-                if($DefaultGateway){ $Parameters["DefaultGateway"] = $DefaultGateway }
+                if($DefaultGateway)
+                {
+                    $Parameters["DefaultGateway"] = $DefaultGateway
+                }
 
                 # If null, remove the gateway from the current settings hash table
-                if($currentSettings['DefaultGateway'] -eq $null){ $currentSettings.remove('DefaultGateway') }
+                if($currentSettings['DefaultGateway'] -eq $null)
+                {
+                    $currentSettings.remove('DefaultGateway')
+                }
 
                 # Remove IP address first; required if the IP is correct but other settings are not
                 Remove-NetIPAddress @currentSettings -confirm:$false -ErrorAction Stop
@@ -232,11 +263,14 @@ function ValidateProperties
                 $null = New-NetIPAddress @Parameters -ErrorAction Stop
 
                 # Make the connection profile private
-                Get-NetConnectionProfile -InterfaceAlias $InterfaceAlias | Set-NetConnectionProfile -NetworkCategory Private -ErrorAction SilentlyContinue
+                Get-NetConnectionProfile -InterfaceAlias $InterfaceAlias | `
+                    Set-NetConnectionProfile -NetworkCategory Private -ErrorAction SilentlyContinue
                 Write-Verbose -Message "IP Interface was set to the desired state."
             }
             else
-            {return $false}
+            {
+                return $false
+            }
         }
         else
         {
@@ -246,8 +280,12 @@ function ValidateProperties
     }
     catch
     {
-       Write-Verbose -Message $_
-       throw "Can not set or find valid IPAddress using InterfaceAlias $InterfaceAlias and AddressFamily $AddressFamily"
+        Write-Verbose -Message $_
+        throw ( @(
+            "Can not set or find valid IPAddress using InterfaceAlias $InterfaceAlias and "
+            "AddressFamily $AddressFamily"
+            ) -join "" | Out-String
+        )
     }
 }
 
--- a/DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1
+++ b/DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1
@@ -143,32 +143,104 @@ function ValidateProperties
         #Get the current IP Address based on the parameters given.
         $currentIP = Get-NetIPAddress -InterfaceAlias $InterfaceAlias -AddressFamily $AddressFamily -ErrorAction Stop
 
+        # Build hash table of current settings
+        $currentSettings = @{}
+        $currentSettings['IPAddress'] = $currentIP.IPAddress
+        $currentSettings['PrefixLength'] = $currentIP.PrefixLength
+        $currentSettings['InterfaceAlias'] = $currentIP.InterfaceAlias
+
+        # Flag to signal whether settings are correct
+        $requiresChanges = $false
+
         #Test if the IP Address passed is equal to the current ip address
-        if(!$currentIP.IPAddress.Contains($IPAddress))
+        if(!$currentSettings['IPAddress'].Equals($IPAddress))
         {
-            Write-Verbose -Message "IPAddress not correct. Expected $IPAddress, actual $($currentIP.IPAddress)"
-            $Parameters = @{}
-
-            #Apply is true in the case of set - target resource - in which case, it will set the new IP Address
-            if($Apply)
-            {
-                Write-Verbose -Message "Setting IPAddress ..."
-                $Parameters["IPAddress"] = $IPAddress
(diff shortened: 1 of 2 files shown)


DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1
```diff
@@ -134,48 +130,162 @@ function ValidateProperties
 
     if(!([System.Net.Ipaddress]::TryParse($ip, [ref]0)))
```
tysonjhayes: This should probably be changed to $IPAddress and the line above it $ip = $IPAddress should just be removed.

DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1
```diff
@@ -134,48 +130,162 @@ function ValidateProperties
...
     {
-       throw "IP Address *$IPAddress* is not in the correct format. Please correct the ipaddress in the configuration and try again"
+        throw ( @(
+            "IP Address *$IPAddress* is not in the correct format. Please correct the ipaddress "
```
tysonjhayes: Sorry Going to nit a little here. We have two different spellings and capitilizations of IPAddress. Could we change this to IPAddress consistently?

DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1
```diff
@@ -134,48 +130,162 @@ function ValidateProperties
...
     try
     {
         Write-Verbose -Message "Checking the IPAddress ..."
         #Get the current IP Address based on the parameters given.
-        $currentIP = Get-NetIPAddress -InterfaceAlias $InterfaceAlias -AddressFamily $AddressFamily -ErrorAction Stop
+        $currentIP = Get-NetIPAddress -InterfaceAlias $InterfaceAlias -AddressFamily `
+            $AddressFamily -ErrorAction Stop
+
+        # Build hash table of current settings
+        $currentSettings = @{}
```
tysonjhayes: nit: Personally I prefer to see hashtables built like this:

````
        $currentSettings = @{
            IPAddress = $currentIP.IPAddress
            PrefixLength = $currentIP.PrefixLength
            InterfaceAlias = $currentIP.InterfaceAlias
        }
````

However, I'm being nitpicky so feel free to ignore me. :grinning:

File: DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1 [hunk 4]
```diff
@@ -134,48 +130,162 @@ function ValidateProperties
...
+            )
+            $requiresChanges = $true
+        }
+        else
+        {
+            Write-Verbose -Message "IPAddress is correct."
+        }
+
+        #Test if the Subnet Mask passed is equal to the current subnet mask
+        if(!$currentSettings['PrefixLength'].Equals([byte]$SubnetMask))
```
tysonjhayes: This should be
````
if (-not $currentSettings['IPAddress'].Equals($IPAddress)) 
````
Which is a bit easier to read.

File: DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1 [hunk 5]
```diff
@@ -134,48 +130,162 @@ function ValidateProperties
...
+        if(!$currentSettings['IPAddress'].Equals($IPAddress))
         {
-            Write-Verbose -Message "IPAddress not correct. Expected $IPAddress, actual $($currentIP.IPAddress)"
-            $Parameters = @{}
+            Write-Verbose -Message ( @(
+                "IPAddress does NOT match desired state. Expected $IPAddress, actual "
+                "$($currentSettings['IPAddress'])."
+                ) -join "" | Out-String
+            )
+            $requiresChanges = $true
```
tysonjhayes: Ok so this is place where I'm going to start arguing breaking up Test and Set. So if I hit this logic I know I need a change. I'm then going to validate PrefixLength, Default Gateway, and DHCP before I actually exit out of this function. If this was purely a Test-TargetResoruce function we could return here and dump out of the script and begin the correction.

DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1
```diff
@@ -134,48 +130,162 @@ function ValidateProperties
...
+        else
+        {
+            Write-Verbose -Message "Subnet mask is correct."
+        }
 
-            #Apply is true in the case of set - target resource - in which case, it will set the new IP Address
+        #Test if the Default Gateway passed is equal to the current default gateway
+        if($DefaultGateway)
+        {
+            $gateway = $DefaultGateway
```
tysonjhayes: Why are we assigning this to a variable when $DefaultGateway should work as is?

DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1
```diff
@@ -134,48 +130,162 @@ function ValidateProperties
...
         #Test if the IP Address passed is equal to the current ip address
-        if(!$currentIP.IPAddress.Contains($IPAddress))
+        if(!$currentSettings['IPAddress'].Equals($IPAddress))
         {
-            Write-Verbose -Message "IPAddress not correct. Expected $IPAddress, actual $($currentIP.IPAddress)"
-            $Parameters = @{}
+            Write-Verbose -Message ( @(
+                "IPAddress does NOT match desired state. Expected $IPAddress, actual "
+                "$($currentSettings['IPAddress'])."
+                ) -join "" | Out-String
```
tysonjhayes: Also I don't know of if the | out-string is actually needed here.

DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1
```diff
@@ -134,48 +130,162 @@ function ValidateProperties
...
+        $currentSettings['IPAddress'] = $currentIP.IPAddress
+        $currentSettings['PrefixLength'] = $currentIP.PrefixLength
+        $currentSettings['InterfaceAlias'] = $currentIP.InterfaceAlias
+
+        # Flag to signal whether settings are correct
+        $requiresChanges = $false
 
         #Test if the IP Address passed is equal to the current ip address
-        if(!$currentIP.IPAddress.Contains($IPAddress))
+        if(!$currentSettings['IPAddress'].Equals($IPAddress))
```
tysonjhayes: Right now this seems to be making the assumption that we're going to have one IP address returned. If I was trying to manage my IPv6 interfaces I have 9 on my desktop right now. This means that I will always fail this check if trying to manage one of those addresses.

What are your thoughts on trying to manage that?

DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1
```diff
@@ -134,48 +130,162 @@ function ValidateProperties
...
+            {
+                throw ( @(
+                    "Default Gateway *$DefaultGateway* is NOT in the correct format. Please "
+                    "correct the gateway ipaddress in the configuration and try again"
+                    ) -join "" | Out-String
+                )
+            }
+
+            $netIpConf = Get-NetIPConfiguration -InterfaceAlias $InterfaceAlias -ErrorAction Stop
+            $currentSettings['DefaultGateway'] = (
```
tysonjhayes: I found this really hard to read what is happening with this.  It took a bit of thinking to realize what properties you were calling and what you were ultimately after.

I'd suggest rewriting 199-202 to look like the following

````
$netIpConfig = Get-NetIPConfiguration -InterfaceAlias $InterfaceAlias -ErrorAction Stop

# IPv6DefaultGateway
# IPv4DefaultGateway
$defaultGatewayProperty = "$($AddressFamily)DefaultGateway"
$currentSettings['DefaultGateway'] = $($netIpConfig.$defaultGatewayProperty).NextHop
````

This will have some comments on what you're creating int the string, and means you are not breaking up the assignment to multiple lines which made it a bit more challenging to understand.

File: DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1 [hunk 10]
```diff
@@ -134,48 +130,162 @@ function ValidateProperties
...
+                    "At least one setting differs from the passed parameters. Applying "
+                    "configuration..."
+                    ) -join "" | Out-String
+                )
+
+                # Build parameter hash table
+                $Parameters = @{}
                 $Parameters["IPAddress"] = $IPAddress
                 $Parameters["PrefixLength"] = $SubnetMask
                 $Parameters["InterfaceAlias"] = $currentIP[0].InterfaceAlias
```
tysonjhayes: So here we are taking the first item in the $CurrentIP array where on line 151 we take all of them. We should probably try to address that.

DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1
```diff
@@ -134,48 +130,162 @@ function ValidateProperties
...
+                if($DefaultGateway)
+                {
+                    $Parameters["DefaultGateway"] = $DefaultGateway
+                }
 
-                if($DefaultGateway){ $Parameters["DefaultGateWay"] = $DefaultGateway }
+                # If null, remove the gateway from the current settings hash table
+                if($currentSettings['DefaultGateway'] -eq $null)
```
tysonjhayes: Is there a scenario where we've specified DefaultGateway and it's equal to null?

So in the current logic $currentSettings wouldn't have a DefaultGateway unless we've actually specified a Default Gateway parameter. Given that we have ValidateNotNullOrEmpty on the parameter, is it possible for us to actually hit this if $DetfaultGateway has been specified?

I ask because this strikes me as it should be:

````
if ($DefaultGateway)
{
    $Parameters["DefaultGateway"] = $DefaultGateway
}
else
{
    $currentSettings.remove('DefaultGateway')
}
````

But I wanted to validate. Probably a decent pester test.

File: DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1 [hunk 12]
```diff
@@ -134,48 +130,162 @@ function ValidateProperties
...
+                Remove-NetIPAddress @currentSettings -confirm:$false -ErrorAction Stop
+
+                # Apply the specified IP configuration
                 $null = New-NetIPAddress @Parameters -ErrorAction Stop
 
                 # Make the connection profile private
-                Get-NetConnectionProfile -InterfaceAlias $InterfaceAlias | Set-NetConnectionProfile -NetworkCategory Private -ErrorAction SilentlyContinue
-                Write-Verbose -Message "IPAddress is set to $IPAddress."
+                Get-NetConnectionProfile -InterfaceAlias $InterfaceAlias | `
+                    Set-NetConnectionProfile -NetworkCategory Private -ErrorAction SilentlyContinue
```
tysonjhayes: The Network Category should probably be specified as a parameter and validated seperately. As I can imagine not wanting to set it to Private.

DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1
```diff
@@ -134,48 +130,162 @@ function ValidateProperties
...
-            Write-Verbose -Message "IPAddress is correct."
+            Write-Verbose -Message "IP Interface is in the desired state."
             return $true
         }
     }
     catch
     {
-       Write-Verbose -Message $_
-       throw "Can not set or find valid IPAddress using InterfaceAlias $InterfaceAlias and AddressFamily $AddressFamily"
+        Write-Verbose -Message $_
```
tysonjhayes: I would caution against throwing your own exception here. If we've run into a problem you are now masking it instead of letting PowerShell throw a more obvious exception.  Honestly I'd probably remove the try/catch and let it fail in its own right considering we'll stop at the failure point if its a bad enough of an exception.

Or at least rethrow whatever the current exception is instead of writing your own.

Tests/MSFT_xIPAddress.Tests.ps1
```diff
@@ -117,7 +139,7 @@ InModuleScope MSFT_xIPAddress {
 
             It 'should be $null' {
                 $Splat = @{
-                    IPAddres = '10.0.0.2'
+                    IPAddress = '10.0.0.2'
                     InterfaceAlias = 'Ethernet'
                 }
                 $Result = ValidateProperties @Splat -Apply
```
tysonjhayes: I would add some calls to make sure the Mocks have been called to prove that its doing more then just running without failure.

Tests/MSFT_xIPAddress.Tests.ps1
```diff
@@ -105,7 +127,7 @@ InModuleScope MSFT_xIPAddress {
 
             It 'should be $true' {
                 $Splat = @{
-                    IPAddres = '192.168.0.1'
+                    IPAddress = '192.168.0.1'
                     InterfaceAlias = 'Ethernet'
                 }
                 $Result = ValidateProperties @Splat
```
tysonjhayes: I'd add in a check that your Mocks were actually called here so we know more then just if it ran without failure.

tysonjhayes:
I'm not going to lie the more I dive into this the more frustrated I am with this resource. Your commits are great in that they are targeted to what you wanted to add in the given context, but it's making me believe a more fundemental rewrite is needed of this resource.

One of the things I would look to address is handling more then one IP address being returned. Even if it's just to throw and bail out.

lorengordon:
Sigh. I was *really* trying to scope my edits just enough to deal with my use case (plus compliance with style guide). Several of your comments are on bits that were in place previously. I can take a stab at a larger rewrite, but I won't be able to get to it for another week or two.

DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1
```diff
@@ -134,48 +130,162 @@ function ValidateProperties
...
+            ) -join "" | Out-String
+        )
     }
 
     try
     {
         Write-Verbose -Message "Checking the IPAddress ..."
         #Get the current IP Address based on the parameters given.
-        $currentIP = Get-NetIPAddress -InterfaceAlias $InterfaceAlias -AddressFamily $AddressFamily -ErrorAction Stop
+        $currentIP = Get-NetIPAddress -InterfaceAlias $InterfaceAlias -AddressFamily `
```
tysonjhayes: Actually to fix the multiple IP addresses that could come back why don't we just add the IPAddress Parameter to this call? This would filter down to the one address that we're looking to manage.

Tests/MSFT_xIPAddress.Tests.ps1
lorengordon: Can you point me in the right direction for how to do that? I'm completely new to the pester tests. I just need a good example that I can copy.

DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1
```diff
@@ -31,9 +29,9 @@ function Get-TargetResource
         [String]$AddressFamily = "IPv4"
     )
 
-
     $returnValue = @{
-        IPAddress = [System.String]::Join(", ",(Get-NetIPAddress -InterfaceAlias $InterfaceAlias -AddressFamily $AddressFamily).IPAddress)
+        IPAddress = [System.String]::Join(", ",(Get-NetIPAddress -InterfaceAlias $InterfaceAlias `
```
tysonjhayes: I'm really unclear why this requests an IP address then doesn't use it for the Get-NetIPAddress call. This could return an array of IP addresses when its unclear that what we expect or not.

DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1
lorengordon: This is testing for the existence of a DefaultGateway in the $currentSettings hash table. If the DefaultGateway is not present, but the parameter is passed to `Remove-NetIPAddress`, the command-let will throw an error. So this check removes the parameter first if the DefaultGateway is not present in the existing configuration.

DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1
lorengordon: Hah, yes, it took me quite a while just to figure out how to get the DefaultGateway at all. It is returned by many of the `Get-Net*` commandlets, far as I could tell. I cut what I figured out and moved on. I'll try to rework it a bit to make it more clear. If you happen to know a more intuitive method of getting the correct DefaultGateway, I'm all ears.

DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1
lorengordon: Oh, *that's* probably why it was `contains` before. For your specific scenario, though, the `Get-NetIPAddress` command is scoped to a single `InterfaceAlias`, which should limit the number of IP addresses returned. Multiple IPs would only be present if multiple IPs were configured on a single IP interface. Possible, of course, so let me do some testing on that one...

DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1
lorengordon: I don't want to break the use case where this resource is trying to change the IP address. Scoping only to the IP address we receive as a parameter will cause this call to return nothing. Which might be fine. I'll look at the subsequent checks and think about it some more.

DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1
lorengordon: I admit to getting quite frustrated trying to figure out the syntax to break a string over multiple lines in a clean manner. It's super easy with python and ''.format(). I tried this approach here and the `-f` method but both require lots of parens, which makes the code rather ugly. When I chanced across something that actually worked, I just ran with it so I could move on. I do feel like I started without the `| out-string` and write-verbose bombed, but I'll try it again and see what happens. Do you happen to know a nicer way to join a multi-line string? I was just trying to keep each line under 100 chars per the style guide (if we figure out how to do this cleanly, it should definitely go into the guide as an example to save others this pain)...

DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1
tysonjhayes: In theory if the IP address doesn't exist are we actually in compliance? In theory we could design this resource to accept one IP and only one IP. If the IP doesn't exist, add it. If we're trying to change the IP we could call this resource twice (once with the Ensure = 'Absent') to remove the IP address we are looking to "change."

tysonjhayes:
I think it's fantastic that you are trying to scope your edits. That's really what we actually want. I'm frustrated at the design of this resource and think it might make more sense to take a step back and rethink it.  You've done good work so far, don't let my comments disuade you of that!

Tests/MSFT_xIPAddress.Tests.ps1
tysonjhayes: [Assert-VerifiableMocks](https://github.com/pester/Pester/wiki/Assert-VerifiableMocks) and [Assert-MockCalled](https://github.com/pester/Pester/wiki/Assert-MockCalled) are the commands.

You can also look at [xWebsite's tests](https://github.com/PowerShell/xWebAdministration/blob/dev/Tests/MSFT_xWebsite.Tests.ps1) as an example of how to do some Pester testing specific to DSC.

DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1
lorengordon: I may be misunderstanding the comment, but this raises and interesting issue with state design. I don't always know the prior state, only what I want the state to be. I'm not necessarily intentionally *changing* the IP, I'm simply specifying the interface configuration I want, and I expect the resource to do all the work to make it so. If I have to know the prior state, then to me there's little benefit to using the stateful functions--I might as well just use the regular PowerShell commandlets.

lorengordon:
I can easily take care of the nits and add tests to make sure the mocks are called. Probably knock that out tomorrow afternoon.

I don't mind taking a stab at a larger rewrite of the resource, but I'd prefer to do that in another pull request, if that's alright with you.

tysonjhayes:
That's great to me, if we're doing a larger rewrite we'll want more eyes on it anyway. :+1:

lorengordon:
Alright, I fixed the nits, added tests to ensure the mocks are called, and I studied the data model for Get-NetIPAddress and Get-NetRoute to try to better support cases where multiple IP addresses or default gateways are present. The behavior currently only supports *managing* a single IP and gateway -- meaning that any other IP or gateway associated with that interface will be removed. However, I added a switch `$NoReset` that is internal only for now (pending the larger rewrite), which we can expose later to support managing multiple IP addresses or gateways on one interface. If `$NoReset=$true`, then other IP addresses or gateways *should* be left alone.

tysonjhayes:
@lorengordon This looks pretty good to me, they've given me merge powers, any objections to merging this? I'm going to open some discussions on where we want to take this resource with some of my ideas early next week so let's hold off on any major rewrites.

lorengordon:
@tysonjhayes, no objections. :shipit: :grin:

lorengordon:
Oh geez, this is embarrassing, I left a few write-verbose statements in there from when I was checking the behavior around the NoReset flag. I'll clean that up really quick...

lorengordon:
Ok, all better.

tysonjhayes:
haha. Totally missed those when I looked. Good catch. :ship:

Issue #20, opened 2015-09-05
Error Assigning both IPv4 and IPv6 DNS addresses to the same Adapter


PlagueHO:
The following DSC config fails:

Configuration MyNetworking {

Import-DscResource -ModuleName xNetworking
Node Computer01 {
	xDnsServerAddress IPv4D2 {
            Address        = '192.168.128.10'
            InterfaceAlias = 'Ethernet 2'
	    AddressFamily  = 'IPv4'
	}
	xDnsServerAddress IPv6D2 {
            Address        = 'fd53:ccc5:895a:bc00::a'
            InterfaceAlias = 'Ethernet 2'
	    AddressFamily  = 'IPv6'
	}
   }
}

With the error:
Test-ConflictingResourcs : A conflict was detected between resources '[xDNSServerAddress]IPv4D2
(C:\VM\BMDLAB.COM\BMDLAB.COM SA_DC1\LabBuilder Files\DSCNetworking.ps1::10::2::xDnsServerAddress)' and
'[xDNSServerAddress]IPv6D2 (C:\VM\BMDLAB.COM\BMDLAB.COM SA_DC1\LabBuilder
Files\DSCNetworking.ps1::22::2::xDnsServerAddress)' in node 'localhost'. Resources have identical key properties but
there are differences in the following non-key properties: 'AddressFamily;Address'. Values 'IPv4;192.168.128.10' don't
match values 'IPv6;fd53:ccc5:895a:bc00::a'. Please update these property values so that they are identical in both
cases.
At line:246 char:9
+         Test-ConflictingResources $keywordName $canonicalizedValue $k ...
+         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [Write-Error], InvalidOperationException
    + FullyQualifiedErrorId : ConflictingDuplicateResource,Test-ConflictingResources
Errors occurred while processing configuration 'Networking'.
At
C:\WINDOWS\system32\WindowsPowerShell\v1.0\Modules\PSDesiredStateConfiguration\PSDesiredStateConfiguration.psm1:3588
char:5
+     throw $ErrorRecord
+     ~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (Networking:String) [], InvalidOperationException
    + FullyQualifiedErrorId : FailToProcessConfiguration

[...]

PR #15, opened 2015-08-26
Adding tests and fixing up Firewall logic

tysonjhayes:
This fixes #12 and adds some integration testing and improves unit testing for xFirewall which helps #14.

I'd appreciate some extra testing on this before I merge as this is likely going to be a breaking change as I changed some of the schema to remove some unneeded parameters and changed some of the parameters to reduce code paths.

PR #21, opened 2015-09-05
Added support for setting DNS for both IPv4 and IPv6 on the same Interface

PlagueHO:
This pull request relates to issue #20. This allows DNS server settings to be set for both IPv4 and IPv6 on the same interface. The change to allow this was the MSFT_xDNSServerAddress/MSFT_xDNSServerAddress.schema.mof by adding a Key attribute to the Address Family:

    [Key,Write,ValueMap{"IPv4", "IPv6"},Values{"IPv4", "IPv6"}] string AddressFamily;

I also added Pester Tests for xDNSServerAddress.


tysonjhayes:
@PlagueHO your change seems pretty good, obviously we'll need to pass the unit tests before I can merge so if you want to fix those up I think we can get this merged.

PlagueHO:
Hi @tysonjhayes ,
Aside from the UTF-8 build issue (didn't load my fork into AppVeyor - sorry my bad), I actually realized there is another issue with this change:
If the AddressFamily is made a [key], it will require the AddressFamily parameter to be changed to mandatory. It isn't currently but it does have a default value. I'm not sure what the policy would be to make such a change as it would mean that anyone upgrading to this new resource would require changes to configs. Is there some other solution to this problem? It does seem likely that assigning IPv4 and IPv6 DNS addresses on the same adapter would need to be supported. Any thoughts on this?

tysonjhayes:
I think I broke the Firewall in PR #15 so, I guess we'll just be breaking a ton of our stuff this release? :grin: 

To your point I'm not sure there is a better way to approach fixing the problem. Fortunately it should find the problem when they go to generate new mof files indicating the key needs to be changed. So it'll throw an error "on compile" (or whatever it's called to make a mof file...) I think it's an acceptable change to support better functionality.

PlagueHO:
Increased functionally is probably the way to go. You're right, an error does occur at Config compile time if AddressFamily is missing so they should be able to resolve the issue then. Cheers.

tysonjhayes:
This looks great, if I could make one suggestion, could we remove the default value from AddressFamily in Get-TargetResource? I could be persuaded to think that we could leave it, just a thought.

Also could you add some notes to README.MD under Version (Create an UNRELEASED header) and note some of the changes. We use that to track the changes we're making until we actually get it released.

PlagueHO:
Hi @tysonjhayes, Absolutely agree about the AddressFamily default - have remove it. Just updating the readme.md now.

tysonjhayes:
@PlagueHO these look good, I see you were changing the outputs to some failure (I can't see what happened in the test pass) and reverted it back, do you want me to merge or do you want to make some more modifications (I'm really good with either).

PlagueHO:
I added the OutputType() attrib to the Test-TargetResource and Get-TargetResource as the DSCHelperTools suggested (to get rid of the warnings), but it cause the tests to fail. I have resolved this in my fork, just testing now. I'll push a new version to this RP in about 10 minutes and you can merge that if you like. Tx

tysonjhayes:
Looks good! Thanks for the work!

PlagueHO:
:+1:

Issue #14, opened 2015-08-21
Improve Test Coverage for all Resources


tysonjhayes:
Currently DNS isn't covered at all that needs to be corrected. Firewall just has some basic tests hitting it. We should get this trued up.

PlagueHO:
I'll throw some tests in for this if you like? I was getting terminating errors thrown by Get-TargetResource when executing Get-DSCConfiguration when the xfirewall resource was being used, so I was going to take a look at this resource anyway and writing pester tests is a good way to get my head around it. It may also give me some ideas on best way to rearchitect xIPAddress to get rid of the Test-Properties helper function - although I don't think there is too much to that really now that the gateway is split out.

tysonjhayes:
Go ye forth @PlagueHO. I'm hoping to loop back these sometime next week and get my firewall fixes working.

PlagueHO:
Hopefully this one can be closed now too as I think test coverage for the existing resources is at about 100% :)

PR #26, opened 2015-09-16
xDefaultGatewayServer resource added

PlagueHO:
This relates to Issue #18 and issue #22. I also did a small amount of work on Issue #14. Documentation has been updated and example usage has been added. I didn't think the Ensure parameter made sense in the xDefaultGatewayAddress resource as just using it without setting an Address parameter (or leaving it blank) will remove the Default Gateway:
```powershell
xDefaultGatewayAddress RemoveDefaultGateway
{
    InterfaceAlias = 'Ethernet'
    AddressFamily  = 'IPv4'
}
```
Setting the default gateway:
```powershell
xDefaultGatewayAddress SetDefaultGateway
{
    Address = '192.168.0.1'
    InterfaceAlias = 'Ethernet'
    AddressFamily  = 'IPv4'
}
```

Other items:
* Corrected the error messages thrown by the xDNSServerAddress resource when an IP Address format and Address Family mismatches.
* Added check to xIPAddress to ensure that IP address format and Address family matches (copied from xDNSServerAddress resource).
* Added some additional tests to MSFT_xIPAddress.Tests.ps1

PR #27, opened 2015-09-19
xFirewall pester tests

PlagueHO:
Relates to Issue #14.
Possibly got a bit test happy here, but you can't have too many tests right? :)

### DIFF
--- a/DSCResources/MSFT_xFirewall/MSFT_xFirewall.psm1
+++ b/DSCResources/MSFT_xFirewall/MSFT_xFirewall.psm1
@@ -463,7 +463,11 @@ function Test-RuleHasProperties
         $desiredConfigurationMatch = $false
     }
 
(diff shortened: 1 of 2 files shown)


tysonjhayes:
I added a bunch of tests in #15 and was planning on working on it today. I'll look between the two of these and figure out if we can combine the two.

PlagueHO:
I should have paid attention to your work in #15. I am taking a look at it now and it is looking pretty good so far. I'll post any comments in that thread.

Just a heads up, I've also made the changes to the xDNSServerAddress, xDefaultGatewayAddress and xIPAddress to break up the logic as per your suggestion in #26. But I'll wait till these existing PRs are sorted before throwing that into the mix :)

tysonjhayes:
Sounds good to me lets get some of #15 in and then we can start burning down the other ones. :+1:

PlagueHO:
I'll close this pull request and re-issue it if there are any tests I can add to the xFirewall tests.

PR #28, opened 2015-09-27
Additional firewall pester tests

PlagueHO:
This PR contains some additional xFirewall Pester tests. They are all for functions not currently covered by the xFirewall.Tests.ps1 module. This relates to Issue #14 . 

When I put these tests in they highlighted a minor logic error in the Test-RuleProperties function - so I fixed that at the same time.


tysonjhayes:
Yeah that is a minor error I made. Good catch. :+1:

Issue #22, opened 2015-09-09
IPAddress requires Gateway configuration


iainbrighton:
I would like to see the enforcement of a default gateway address when setting an IP address (also discussed in issue #18) removed. The gateway address really has nothing to do with the IP address assignment other than MS putting it in the same GUI!

This should be a relatively easy, non-breaking change too.

tysonjhayes:
This is fixed with #26. Thanks!

Issue #25, opened 2015-09-15
xFirewall: Need a way of enabling built-in Firewall Rule without setting all properties


PlagueHO:
An example:
I was trying to use xFirewall to enable the built-in Firewall rules for FSRM:
![before](https://cloud.githubusercontent.com/assets/7589164/9892704/70499db4-5c65-11e5-9554-4cdceb6cdf83.png)
Using:
```powershell
xFirewall FSRMFirewall8
{
	Name = "FSRM-SrmSvc-In (RPC)"
	State = "Enabled"
	Access = "Allow"
}
```

It was enabling the rule just fine, but was replacing the properties:
![after](https://cloud.githubusercontent.com/assets/7589164/9892740/e4f23130-5c65-11e5-867d-90f30556dab2.png)

This can be worked around by manually setting all the properties of the Firewall rule but it is a bit of a pain, especially as the built in rules contain a lot of settings and there are often a lot of them - e.g. 8 rules for FSRM.

It would be nice if this would enable a built-in rule:
xFirewall FSRMFirewall8
{
	Name = "FSRM-SrmSvc-In (RPC)"
	State = "Enabled"
}

It would be possible to implement a special code path for when only the Name and State are passed and the rule exists that just fires:
Enable-NetFirewallRule or Disable-NetFirewallRule

Or is there a better way of doing this? I'm not completely sure that this would be a 100% obvious way of using the xFirewall resource because if any other parameter was set with State it would revert to the Delete and Recreate method which would still remove all properties of the built-in rule. Perhaps an new resource could be created for just disabling/enabling existing firewall rules?

PlagueHO:
I've been thinking about this issue and I think I can see a way the resource could be changed to support this user story fairly easily.

The issue at the moment is in lines 131-134 of MSFT_xFirewall.psm1:
```powershell
Remove-NetFirewallRule -Name $Name

# Set the Firewall rule based on specified parameters
New-NetFirewallRule @PSBoundParameters
```
This code fires whenever a rule needs to be updated. This effectively destroys the structure of a built in rule.

What could be done instead is to build a @splat object out of the PSBoundParameters and then modify the splat object adding, removing or tweaking any parameters required and then splatting it on Set-NetFirewallRule (removing the Remove/New).

Pretty much just this would do (perhaps some tweaking would be required):
```powershell
$Splat = $PSBoundParameters
Set-NetFirewallRule @Splat
```

[...]

Issue #19, opened 2015-09-03
Create NetConnectionProfile resource


tysonjhayes:
I'm thinking we should create a NetConnecitonProfile resource. This would allow the management of the NetworkCategories  of Private/Public/Domain etc on the interface level. This would also mean pulling it it out of xIPAddress which would solve #3 by doing this.

Before doing this work I wanted to ask the community their thoughts on the design.

Thanks!

Dan1el42:
Great idea :-). I actually have a draft of an DSC resource call xNetConnectionProfile on my cloud drive.

The basic design I would like to propose is:

Get/Set/Test-TargetResource invoke internal functions Get/Set/Test-NetConnectionProfileInternal which invoke the Get/Set-NetConnectionProfile functions of the PowerShell module included with each OS with some additional logic and verbose logging.

I think the advantages of the internal functions are easier debugging and testing without the LCM, and the ability to use the internal functions in a future versions of the resource module with a DscResource class instead.

Get/Set/Test-TargetResource do not need extensive tests because each contains basically one line of code splatting the parameters to the internal functions and returning an object if applicable. The primary testing would be happening on the internal functions with mocks for Get/Set-NetConnectionProfile functions for Unit tests.

Thoughts?

P.S. I think I can make the time this coming Saturday to complete the resource and submit it as pull request.

Best,
Daniel

tysonjhayes:
I'm not really a fan of internal functions unless they serve a good purpose, just my opinion really, just seems an unnecessary level of complexity when the three primary funcitons work just as well. However, I can appreciate the thought process behind the idea and will withold judgement until I can actually see it. :grinning: 

If you have time this Saturday to take a crack at it I'd love to see it!

iainbrighton:
FYI - I created this a while ago https://github.com/iainbrighton/cNetworkProfile. I'm happy for it to be developed further, merged into another repo or completely ignored. Some of the code might help?

Dan1el42:
@iainbrighton, Thanks for the pointer. I will checkout your resources and look into how we can merge them.

PR #29, opened 2015-09-29
Rearchitect xIPAddress, xDNSServerAddress and xDefaultGatewayAddress

PlagueHO:
The resources were modified in the following ways:
1. Test-Properties function removed and logic moved into Test-TargetResource and/or Set-TargetResource so that code to test state is not called twice. As per recommendation from @tysonjhayes in #22.
2. Parameter test logic removed into Test-ResourceProperties function with additional checking for IPv6 parameters.
3. Additional Pester tests added to each resource - including some IPv6 specific tests.
4. "Lightbulb" changes suggested by ISESteroids interface.
5. Attempt to mimic code style of new xFirewall resource.
Code reviews gratefully accepted :)


Tests/Unit/MSFT_xIPAddress.Tests.ps1
```diff
@@ -119,102 +178,246 @@ InModuleScope MSFT_xIPAddress {
...
+                    AddressFamily = 'IPv6'
+                }
+                { Test-ResourceProperty @Splat } | Should Not Throw
             }
         }
     }
 }
 
-# Cleanup after the test
-Remove-Item -Path $moduleRoot -Recurse -Force
```
tysonjhayes: Any reason you decided to remove this?

tysonjhayes:
Gave it a quick glance, looks good so far, going to block some time in a few hours to go into this in detail. Thanks for the work!

Tests/Unit/MSFT_xIPAddress.Tests.ps1
PlagueHO: Actually, there isn't any reason! I think what might have happened is that I copied the header code xFirewall.Tests code you added in the recent version but didn't copy the footer cleanup code and therefore when I committed it removed it. I'll add it to mine as it should be there - always best to clean up afterwards :)

DSCResources/MSFT_xDNSServerAddress/MSFT_xDNSServerAddress.psm1
```diff
@@ -55,11 +59,52 @@ function Set-TargetResource
...
+        #Should not need to validate the settings because Test-TargetResource already did this
+
+        #Get the current DNS Server Addresses based on the parameters given.
+        $currentAddress = (Get-DnsClientServerAddress `
+            -InterfaceAlias $InterfaceAlias `
+            -AddressFamily $AddressFamily `
+            -ErrorAction Stop).ServerAddresses
+
+        #Check if the Server addresses are the same as the desired addresses.
+        if(@(Compare-Object `
```
tysonjhayes: nit: Could we put a space between if and the parantheses?

DSCResources/MSFT_xDNSServerAddress/MSFT_xDNSServerAddress.psm1
```diff
@@ -55,11 +59,52 @@ function Set-TargetResource
...
     )
 
-    Test-Properties @PSBoundParameters -Apply
+    try
+    {        
+        Write-Verbose -Message ( @("$($MyInvocation.MyCommand): "
+            'Applying the DNS Server Address ...'
+            ) -join '')
+
+        #Should not need to validate the settings because Test-TargetResource already did this
```
tysonjhayes: While this comment makes sense in this review, I'm not sure it actually adds any value in the long term. Could we remove it?

DSCResources/MSFT_xDNSServerAddress/MSFT_xDNSServerAddress.psm1
```diff
@@ -80,93 +125,124 @@ function Test-TargetResource
...
-    {        
-        Write-Verbose -Message "Checking the DNS Server Address ..."
-        #Get the current IP Address based on the parameters given.
-        $currentAddress = (Get-DnsClientServerAddress -InterfaceAlias $InterfaceAlias -AddressFamily $AddressFamily -ErrorAction Stop).ServerAddresses
-
-        #Check if the Server addresses are the same as the desired addresses.
-        if(@(Compare-Object -ReferenceObject $currentAddress -DifferenceObject $Address -SyncWindow 0).Length -gt 0)
+    if(-not (Get-NetAdapter | Where-Object -Property Name -EQ $InterfaceAlias ))
+    {
+                throw ( @(
```
tysonjhayes: Could we move this to the left by two tabs? It looks kind of off being this far off.

DSCResources/MSFT_xDNSServerAddress/MSFT_xDNSServerAddress.psm1
```diff
@@ -80,93 +125,124 @@ function Test-TargetResource
...
+                    "Interface $InterfaceAlias is not available. "
+                    'Please select a valid interface and try again'
+                    ) -join '' )
+    }
+    if ($Address)
```
tysonjhayes: Is there a situation where we are going to be passing in an address? If not we can move this entire thing up a level.

DSCResources/MSFT_xDNSServerAddress/MSFT_xDNSServerAddress.psm1
```diff
@@ -80,93 +125,124 @@ function Test-TargetResource
...
+    {
+        if(-not ([System.Net.Ipaddress]::TryParse($Address, [ref]0)))
+        {
+                throw ( @(
```
tysonjhayes: Could we move this block to the left by a tab please?

DSCResources/MSFT_xDNSServerAddress/MSFT_xDNSServerAddress.psm1
```diff
@@ -80,93 +125,124 @@ function Test-TargetResource
...
-# "Apply" is set, then it will set the properties after a test
+#  Helper functions
 #######################################################################################
-function Test-Properties
-{
+function Test-ResourceProperty {
+    # Function will check the Address details are valid and do not conflict with
+    # Address family. Ensures interface exists.
+    # If any problems are detected an exception will be thrown.
+    [CmdLetBinding()]
```
tysonjhayes: Could you change this to CmdletBinding (not CmdLetBinding) please?

DSCResources/MSFT_xDNSServerAddress/MSFT_xDNSServerAddress.psm1
```diff
@@ -80,93 +125,124 @@ function Test-TargetResource
...
+    {
+        if(-not ([System.Net.Ipaddress]::TryParse($Address, [ref]0)))
+        {
+                throw ( @(
+                    "Address $Address is not in the correct format. "
+                    'Please correct the Address parameter in the configuration and try again.'
+                    ) -join ''
+                )
+        }
+        if (([System.Net.IPAddress]$Address).AddressFamily.ToString() `
```
tysonjhayes: Suggestion: you could change up this entire chunk to the following which would reduce some of the logic nesting that you currently have while maintaining the basic premise.

````powershell
function Test-ResourceProperty
{
    # Function will check the Address details are valid and do not conflict with
    # Address family. Ensures interface exists.
    # If any problems are detected an exception will be thrown.
    [CmdletBinding()]
    param
    (
        [String]$Address,

        [Parameter(Mandatory)]
        [ValidateNotNullOrEmpty()]
        [String]$InterfaceAlias,

        [ValidateSet('IPv4', 'IPv6')]
        [String]$AddressFamily = 'IPv4'
    )

    if( -not (Get-NetAdapter | Where-Object -Property Name -EQ $InterfaceAlias ))
    {
        throw ( @(
            "Interface $InterfaceAlias is not available. "
            'Please select a valid interface and try again'
            ) -join '' )
    }

    if( -not ([System.Net.Ipaddress]::TryParse($Address, [ref]0)))
    {
        throw ( @(
            "Address $Address is not in the correct format. "
            'Please correct the Address parameter in the configuration and try again.'
            ) -join ''
        )
    }

    if (([System.Net.IPAddress]$Address).AddressFamily.ToString() `
        -eq [System.Net.Sockets.AddressFamily]::InterNetwork.ToString() -and $AddressFamily -ne 'IPv4')
    {
        throw ( @(
                "Address $Address is in IPv4 format, which does not match "
                "server address family $AddressFamily. "
                'Please correct either of them in the configuration and try again.'
            ) -join '' )
    }

    if ($AddressFamily -ne 'IPv6')
    {
        throw ( @(
                "Address $Address is in IPv6 format, which does not match "
                "server address family $AddressFamily. "
                'Please correct either of them in the configuration and try again'
            ) -join '' )
    }
} # Test-ResourceProperty
````

File: DSCResources/MSFT_xDNSServerAddress/MSFT_xDNSServerAddress.psm1 [hunk 9]
```diff
@@ -80,93 +125,124 @@ function Test-TargetResource
...
+        }
+        else 
+        { 
+            #Test will return true in this case
+            Write-Verbose -Message ( @( "$($MyInvocation.MyCommand): "
+                'DNS Servers are set correctly.' 
+                ) -join '' )
+        }
+    }
+    catch
```
tysonjhayes: Doesn't this swallow any exception you threw in Test-ResourceProperty considering the throw (in Test-ResourceProperty) is happening within your try/catch on this level?

Given that the helper function is expected to throw do we really need the try/catch here?

DSCResources/MSFT_xDNSServerAddress/MSFT_xDNSServerAddress.psm1
```diff
@@ -80,93 +125,124 @@ function Test-TargetResource
...
+        }
 
+        #Get the current DNS Server Addresses based on the parameters given.
+        $currentAddress = (Get-DnsClientServerAddress `
+            -InterfaceAlias $InterfaceAlias `
+            -AddressFamily $AddressFamily `
+            -ErrorAction Stop).ServerAddresses
+
+        #Check if the Server addresses are the same as the desired addresses.
+        if(@(Compare-Object `
```
tysonjhayes: nit: I found this a little hard to read, could we assign this to a variable and if/else on that?

ie: 

````powershell
        $addressCompare = (Compare-Object -ReferenceObject $currentAddress `
            -DifferenceObject $Address -SyncWindow 0).Length -gt 0

        if ($addressCompare)
````

File: DSCResources/MSFT_xDNSServerAddress/MSFT_xDNSServerAddress.psm1 [hunk 11]
```diff
@@ -80,93 +125,124 @@ function Test-TargetResource
...
+    }
+    catch
+    {
+        Write-Verbose -Message ( @( "$($MyInvocation.MyCommand): "
+            "Error testing valid DNS Server addresses using InterfaceAlias $InterfaceAlias " 
+            "and AddressFamily $AddressFamily"
+            ) -join '' )
+        throw $_.Exception
+    }
+    return -not $requiresChanges
```
tysonjhayes: I found this a bit confusing at first, we're returning the opposite of whatever this is?

Could we change the variable name to something like $DesiredConfigurationMatch = $true and change to false if address compare fails? That would be a bit easier to read for me.

DSCResources/MSFT_xDNSServerAddress/MSFT_xDNSServerAddress.psm1
tysonjhayes: Actually on second thought, why do we need this? If Test-TargetResource indicates we need to make a change why are we again seeing if it matches?

In fact I would also argue that we don't even need the try/catch in the set-targetresource. Set-DnsClientServerAddress should have its own erorr messaging that would probably clearer to resolve through research then the error message we're obscuring by throwing our own.  If we were to do that then the entire function would consist of the messaging around begining, and ending and applying the change. This would greatly reduce the logic and testing area on the function.

DSCResources/MSFT_xDNSServerAddress/MSFT_xDNSServerAddress.psm1
```diff
@@ -55,11 +59,52 @@ function Set-TargetResource
...
+    try
+    {        
+        Write-Verbose -Message ( @("$($MyInvocation.MyCommand): "
+            'Applying the DNS Server Address ...'
+            ) -join '')
+
+        #Should not need to validate the settings because Test-TargetResource already did this
+
+        #Get the current DNS Server Addresses based on the parameters given.
+        $currentAddress = (Get-DnsClientServerAddress `
```
tysonjhayes: Could we use the ````@PSBoundParameters```` here?

DSCResources/MSFT_xDefaultGatewayAddress/MSFT_xDefaultGatewayAddress.psm1
```diff
@@ -187,77 +212,92 @@ function Test-Properties
...
+        [String]$AddressFamily = 'IPv4'
+    )
 
-                Write-Verbose -Message "Default Gateway address was set to the desired state."
-                return $true
-            }
-            else
+    if(-not (Get-NetAdapter | Where-Object -Property Name -EQ $InterfaceAlias ))
+    {
+                throw ( @(
```
tysonjhayes: Could you move this to the left by two tabs please?

DSCResources/MSFT_xDefaultGatewayAddress/MSFT_xDefaultGatewayAddress.psm1
```diff
@@ -187,77 +212,92 @@ function Test-Properties
...
+                    "Interface $InterfaceAlias is not available. "
+                    'Please select a valid interface and try again'
+                    ) -join '')
+    }
+    if ($Address)
+    {
+        if(-not ([System.Net.Ipaddress]::TryParse($Address, [ref]0)))
+        {
+                throw ( @(
```
tysonjhayes: Could you move this to the left by two tabs please?

DSCResources/MSFT_xDefaultGatewayAddress/MSFT_xDefaultGatewayAddress.psm1
```diff
@@ -187,77 +212,92 @@ function Test-Properties
...
-            {
-                Write-Verbose -Message ( @(
-                    "At least one setting differs from the passed parameters. Applying "
-                    "configuration..."
-                    ) -join ""
-                )
+#######################################################################################
+#  Helper functions
+#######################################################################################
+function Test-ResourceProperty {
```
tysonjhayes: Given this function is pretty much a copy/paste of xDNSServerAddress I'm going to refer my comments about it to that one. :grin:

DSCResources/MSFT_xDefaultGatewayAddress/MSFT_xDefaultGatewayAddress.psm1
```diff
@@ -187,77 +212,92 @@ function Test-Properties
...
             else
             {
-                Write-Verbose -Message "Default Gateway does not exist which is correct."
+                Write-Verbose -Message ( @("$($MyInvocation.MyCommand): "
+                    'Default Gateway does not exist which is correct.'
+                    ) -join '' )
             }
         }
+    }
+    catch
```
tysonjhayes: Same comment here that I said about xDNSServerAddress is this really needed given Test-ResourceProperty is supposed to throw?

DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1
```diff
@@ -56,12 +61,92 @@ function Set-TargetResource
...
+
+        # Use $AddressFamily to select the IPv4 or IPv6 destination prefix
+        $DestinationPrefix = '0.0.0.0/0'
+        if($AddressFamily -eq 'IPv6')
+        {
+            $DestinationPrefix = '::/0'
+        }
+
+        # Get all the default routes - this has to be done in case the IP Address is
+        # beng Removed
```
tysonjhayes: being removed

DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1
```diff
@@ -56,12 +61,92 @@ function Set-TargetResource
...
+            PrefixLength = $SubnetMask
+            InterfaceAlias = $InterfaceAlias
+        }
+
+        # Apply the specified IP configuration
+        $null = New-NetIPAddress @Parameters -ErrorAction Stop
+
+        # Make the connection profile private
+        Get-NetConnectionProfile -InterfaceAlias $InterfaceAlias | `
+            Set-NetConnectionProfile -NetworkCategory Private -ErrorAction SilentlyContinue
```
tysonjhayes: I know this is the current logic in the system, but I believe we specified #19 requested that we set this else where. Could we remove the modification of the profile or do we have an issue if this isn't set after we've modified the IP address?

DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1
```diff
@@ -56,12 +61,92 @@ function Set-TargetResource
...
+        Write-Verbose -Message ( @("$($MyInvocation.MyCommand): "
+            'IP Interface was set to the desired state.'
+            ) -join '' )
+    }
+    catch
```
tysonjhayes: Not sure try/catch is really needed here. If one of these functions throw it may present a better error message then we are presenting.

DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1
```diff
@@ -185,102 +228,108 @@ function Test-Properties
...
+                ) -join '' )
         }
+    } catch {
+        Write-Verbose -Message ( @( "$($MyInvocation.MyCommand): "
+            "Error testing valid IPAddress using InterfaceAlias $InterfaceAlias "
+            "and AddressFamily $AddressFamily"
+            ) -join '' )
+        throw $_.Exception
+    }
+    return -not $requiresChanges
```
tysonjhayes: Again can we change this?

DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1
```diff
@@ -185,102 +228,108 @@ function Test-Properties
...
             $requiresChanges = $true
         }
         else
         {
-            Write-Verbose -Message "DHCP is already disabled."
+            Write-Verbose -Message ( @( "$($MyInvocation.MyCommand): "
+                'DHCP is already disabled.'
+                ) -join '' )
         }
+    } catch {
```
tysonjhayes: Do we really think try/catch is needed here or is the other places we throw better?

DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1
```diff
@@ -185,102 +228,108 @@ function Test-Properties
...
+    }
+    if(-not ([System.Net.Ipaddress]::TryParse($IPAddress, [ref]0)))
+    {
+            throw ( @(
+                "IP Address $IPAddress is not in the correct format. "
+                'Please correct the IPAddress parameter in the configuration and try again.'
+                ) -join ''
+            )
+    }
+    if (([System.Net.IPAddress]$IPAddress).AddressFamily.ToString() `
```
tysonjhayes: See comments I've made about this previously.

tysonjhayes:
OK - I'm going to hold off making comments about the unit tests until we've addressed some of my other comments as if we take them would change what we expect the tests to do. :laughing: So no sense in spending to much time on them right now.

Overall I think these are good changes, I think we can make some improvements on the logic but this is exactly what I was looking for in a refactor for these functions. Thanks a bunch!

DSCResources/MSFT_xDNSServerAddress/MSFT_xDNSServerAddress.psm1
PlagueHO: :+1: I think the try/catch in the Test-TargetResource is really unnecessary, so I've gotten rid of it in xIPAddress,xDNSServerAddress and xDefaultGatewayAddress.

DSCResources/MSFT_xDNSServerAddress/MSFT_xDNSServerAddress.psm1
PlagueHO: :+1: Done in all 3 resources.

DSCResources/MSFT_xDNSServerAddress/MSFT_xDNSServerAddress.psm1
PlagueHO: Good stuff :) I replaced the function with this one.

DSCResources/MSFT_xDefaultGatewayAddress/MSFT_xDefaultGatewayAddress.psm1
PlagueHO: :+1:  All your previous comments I've applied to this resource (as well as xDefaultGatewayAddress) :)

DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1
PlagueHO: I'll remove it as I think #19 should be implemented reasonably quickly (I'm happy to put code in for that later on this week once this PR is done).

PlagueHO:
OK, I think I've made all the changes as per your comments (hopefully I didn't miss any :) ). It also highlighted a couple of logic problems in the Test-ResourceProperty function (and Pester tests) that I have fixed, mainly relating to ensuring IPv6 addresses are tested correctly. I tried to make sure any comment on one resource was reflected in all the other resources as I try to go for consistency. :)

Great to get to work on these :) Cheers.

Tests/Unit/MSFT_xDefaultGatewayAddress.Tests.ps1
```diff
@@ -50,193 +80,227 @@ InModuleScope MSFT_xDefaultGatewayAddress {
...
                 $Splat = @{
-                    Apply = $true
                     Address = '192.168.0.1'
-                    InterfaceAlias = 'Ethernet'
+                    InterfaceAlias = 'NotReal'
                     AddressFamily = 'IPv4'
                 }
-                $Result = Test-Properties @Splat
-                $Result | Should Be $true
+                { Test-ResourceProperty @Splat } | Should Throw
```
tysonjhayes: We should be more specific on what is expected to throw. 

So this would be

````powershell
$errorRecord = "Interface NotReal is not available. Please select a valid interface and try again"
{ Test-ResourceProperty @Splat } | Should Throw $errorRecord
````

That being said we may want to use a string hashtable at the top of the module to allow us to keep all the strings in one place and call them into the tests.  You can see examples of the string hash [here] (https://github.com/PowerShell/xWebAdministration/blob/dev/DSCResources/MSFT_xWebsite/MSFT_xWebsite.psm1#L5) and the resulting test calling it [here] (https://github.com/PowerShell/xWebAdministration/blob/dev/Tests/Unit/MSFT_xWebsite.Tests.ps1#L266).  This way we can update the strings in a central location and localize them if we ever need to.

This also could totally be done as a seperate PR, just makes testing the error messages easier. :+1:

File: Tests/Unit/MSFT_xDefaultGatewayAddress.Tests.ps1 [hunk 24]
```diff
@@ -50,193 +80,227 @@ InModuleScope MSFT_xDefaultGatewayAddress {
...
-
-            It 'should call Get-NetRoute three times' {
-                Assert-MockCalled -commandName Get-NetRoute -Exactly 3
-            }
-            It 'should call Remove-NetRoute two times' {
-                Assert-MockCalled -commandName Remove-NetRoute -Exactly 2
-            }
-            It 'should call New-NetRoute once' {
-                Assert-MockCalled -commandName New-NetRoute -Exactly 1
+                { Test-ResourceProperty @Splat } | Should Throw
```
tysonjhayes: We should specify the error message here as well.

Tests/Unit/MSFT_xDefaultGatewayAddress.Tests.ps1
```diff
@@ -50,193 +80,227 @@ InModuleScope MSFT_xDefaultGatewayAddress {
...
                 $Splat = @{
-                    Apply = $True
+                    Address = '192.168.0.1'
                     InterfaceAlias = 'Ethernet'
-                    AddressFamily = 'IPv4'
+                    AddressFamily = 'IPv6'
                 }
-                $Result = Test-Properties @Splat
-                $Result | Should Be $true
+                { Test-ResourceProperty @Splat } | Should Throw
```
tysonjhayes: And here.

Tests/Unit/MSFT_xDefaultGatewayAddress.Tests.ps1
```diff
@@ -50,193 +80,227 @@ InModuleScope MSFT_xDefaultGatewayAddress {
...
+            It 'should not throw an error' {
                 $Splat = @{
-                    Apply = $True
+                    Address = '192.168.0.1'
                     InterfaceAlias = 'Ethernet'
                     AddressFamily = 'IPv4'
                 }
-                $Result = Test-Properties @Splat
-                $Result | Should Be $true
+                { Test-ResourceProperty @Splat } | Should Not Throw
```
tysonjhayes: ...and well, you get the idea. :grinning:

Tests/Unit/MSFT_xDefaultGatewayAddress.Tests.ps1
PlagueHO: Ah good stuff! Didn't know you could do that (specify the error message expected from the throw). The hash table makes a lot of sense as well. I'll get it done in this PR later on today! Thanks heaps for the feedback - very helpful!

PlagueHO:
Hi @tysonjhayes - I've made the modifications to the xDefaultGatewayAddress resource, but before I go putting the same changes into the other resources I thought I'd run this one past you to make sure it was going in the right direction :) I pretty much replicated the method/code from the xWebAdministration resources you linked to.

I also refined the try/catch blocks in the Set-TargetResource so that there is no all encompassing try/catch, instead there are individual try/catches round the actual cmdlets that do the work so the resource provides more specific feed back when errors occur.

If you're OK with the changes to this resource, I'll replicate them in the others.

DSCResources/MSFT_xDefaultGatewayAddress/MSFT_xDefaultGatewayAddress.psm1
```diff
@@ -21,15 +43,19 @@ function Get-TargetResource
...
-    $DestinationPrefix = "0.0.0.0/0"
-    if($AddressFamily -eq "IPv6")
+    $DestinationPrefix = '0.0.0.0/0'
+    if ($AddressFamily -eq 'IPv6')
     {
-        $DestinationPrefix = "::/0"
+        $DestinationPrefix = '::/0'
     }
     # Get all the default routes
     $defaultRoutes = Get-NetRoute -InterfaceAlias $InterfaceAlias -AddressFamily `
```
tysonjhayes: nit: Line 63 should be Where-Object

tysonjhayes:
Forgive me as I'm going over Gateway with a fine tooth comb, currently this only returns Address if it finds a value, given null is an acceptable value of a Default Gateway (in this case it would be no gateway defined) wouldn't it make sense to return null as the value in Get-TargetResource? Just a thought. :+1:

Tests/Unit/MSFT_xDefaultGatewayAddress.Tests.ps1
```diff
@@ -50,193 +80,259 @@ InModuleScope MSFT_xDefaultGatewayAddress {
...
-            It 'when default gateway matches should be $true' {
-                $Splat = @{
-                    Address = '192.168.0.1'
-                    InterfaceAlias = 'Ethernet'
-                    AddressFamily = 'IPv4'
-                }
-                $Result = Test-Properties @Splat
-                $Result | Should Be $true
+            It 'should call all the mocks' {
+                Assert-MockCalled -commandName Get-NetAdapter -Exactly 0
```
tysonjhayes: Get-NetAdapter isn't supposed to be in set-targetresource, why are we checking it wasn't called?

Tests/Unit/MSFT_xDefaultGatewayAddress.Tests.ps1
```diff
@@ -50,193 +80,259 @@ InModuleScope MSFT_xDefaultGatewayAddress {
...
+                Assert-MockCalled -commandName Get-NetRoute -Exactly 1
+                Assert-MockCalled -commandName Remove-NetRoute -Exactly 1
             }
+        }
 
-            It 'when default gateway is not passed should be $false' {
+        Context 'invoking with valid Default Gateway Address' {
+            It 'should rerturn $null' {
```
tysonjhayes: return

Tests/Unit/MSFT_xDefaultGatewayAddress.Tests.ps1
```diff
@@ -50,193 +80,259 @@ InModuleScope MSFT_xDefaultGatewayAddress {
...
+
+        Mock Remove-NetRoute
 
-            It 'when default gateway does not match should be $false' {
+        Mock New-NetRoute
+
+        #endregion
+
+        Context 'invoking with no Default Gateway Address' {
+            It 'should rerturn $null' {
```
tysonjhayes: return

Tests/Unit/MSFT_xDefaultGatewayAddress.Tests.ps1
```diff
@@ -50,193 +80,259 @@ InModuleScope MSFT_xDefaultGatewayAddress {
...
-                    Address = '192.168.0.1'
-                    InterfaceAlias = 'Ethernet'
-                    AddressFamily = 'IPv4'
-                }
-                $Result = Test-Properties @Splat
-                $Result | Should Be $true
+            It 'should call all the mocks' {
+                Assert-MockCalled -commandName Get-NetAdapter -Exactly 0
+                Assert-MockCalled -commandName Get-NetRoute -Exactly 1
+                Assert-MockCalled -commandName Remove-NetRoute -Exactly 1
```
tysonjhayes: It may make sense to do a check that New-NetRoute wasn't called here to make sure we weren't hitting the $Address logic.

Tests/Unit/MSFT_xDefaultGatewayAddress.Tests.ps1
PlagueHO: Good catch! I used to call test-resourceproperty in this function, but removed it. Didn't remove the mock. Have now :)

Tests/Unit/MSFT_xDefaultGatewayAddress.Tests.ps1
PlagueHO: Great eye there!

tysonjhayes: Thanks, helps that we've scopped this one down a bit more so its easier to comb over. :laughing:

Tests/Unit/MSFT_xDefaultGatewayAddress.Tests.ps1
PlagueHO: All done :) Thank again for the feedback - really appreciated. Shall I go ahead and mimic the changes to the xDNSServerAddress and xIPAddress resources?

Tests/Unit/MSFT_xDefaultGatewayAddress.Tests.ps1
```diff
@@ -50,193 +80,259 @@ InModuleScope MSFT_xDefaultGatewayAddress {
...
+        Context 'checking return with default gateway but none are currently set' {
+            It 'should return false' {
+
+                $Splat = @{
+                    Address = '192.168.0.1'
                     InterfaceAlias = 'Ethernet'
-                    InterfaceIndex = 1
                     AddressFamily = 'IPv4'
                 }
+                Test-TargetResource @Splat | Should Be $False
```
tysonjhayes: This is probably going overboard, but there is ways to capture verbose output, it could be interesting to test that the right verbose streams were used to make sure the right logic was hit. maybe. lol.

DSCResources/MSFT_xDNSServerAddress/MSFT_xDNSServerAddress.psm1
```diff
@@ -80,93 +132,110 @@ function Test-TargetResource
...
+    if ( -not (Get-NetAdapter | Where-Object -Property Name -EQ $InterfaceAlias ))
+    {
+        throw ( @(
+            "Interface $InterfaceAlias is not available. "
+            'Please select a valid interface and try again'
+            ) -join '' )
     }
-    catch
+
+    if ( -not ([System.Net.Ipaddress]::TryParse($Address, [ref]0)))
```
tysonjhayes: nit: Capitlization should be IPAddress

DSCResources/MSFT_xDefaultGatewayAddress/MSFT_xDefaultGatewayAddress.psm1
```diff
@@ -89,175 +212,171 @@ function Test-TargetResource
...
+            -ArgumentList $errorMessage
+        $errorRecord = New-Object -TypeName System.Management.Automation.ErrorRecord `
+            -ArgumentList $exception, $errorId, $errorCategory, $null
+
+        $PSCmdlet.ThrowTerminatingError($errorRecord)
     }
-}
+    if ($Address)
+    {
+        if (-not ([System.Net.Ipaddress]::TryParse($Address, [ref]0)))
```
tysonjhayes: nit: Capitilzation should be IPAddress

DSCResources/MSFT_xDefaultGatewayAddress/MSFT_xDefaultGatewayAddress.psm1
```diff
@@ -89,175 +212,171 @@ function Test-TargetResource
...
+            $errorId = 'AddressFormatError'
+            $errorCategory = [System.Management.Automation.ErrorCategory]::InvalidArgument
+            $errorMessage = $($LocalizedData.AddressFormatError) -f $Address
+            $exception = New-Object -TypeName System.InvalidOperationException `
+                -ArgumentList $errorMessage
+            $errorRecord = New-Object -TypeName System.Management.Automation.ErrorRecord `
+                -ArgumentList $exception, $errorId, $errorCategory, $null
+
+            $PSCmdlet.ThrowTerminatingError($errorRecord)
+        }
```
tysonjhayes: nit: could we add a line break here and after $detectedAddressFamily it'd make it a bit easier to read.

Tests/Unit/MSFT_xDefaultGatewayAddress.Tests.ps1
PlagueHO: Haha, the thought of checking the verbose output had crossed my mind once or twice as a way of checking certain scenarios, but I generally just went with ensuring the mocks were called. But could be something to implement if bugs slip through :)

tysonjhayes: yeah as I said, probably going overboard, so let's skip it for now. :+1:

tysonjhayes:
So my only thoughts at this point continue to be around the use of try/catch, if its needed, and if we should be throwing our own errors.

Specifically the way I'm thinking about it as such:
````powershell
try
{
    throw 'some easily found error on bingoogle/stack overflow'
}
catch
{
    throw 'an exception that only occurs in our code but we didn't actually write the function'
}
````

With the example I gave, if Get-NetRoute throws an exception it'll be a "known" exception that a user could find other examples of, by rethrowing the exception we'll lose that information and may obscure resolution of the issue as debugging can be quite painful on these scripts (source: I've done it while developing a bunch of these).

I'm not saying we need to change it, just something that keeps coming to mind while I look at this code.

As it stands this looks good to continue forward with to the rest of the resources, thanks for all of the hard work!

PlagueHO:
I completely get what you're saying - why bother wrap any miscellaneous errors with a try/catch as we're just obfuscating the original error. I'd bow to your experience here and say that I should remove the try/catches around these specific cmdlets and just make sure they've got -ErrorAction Stop parameter on each cmdlet - just let any cmdlet errors be handled natively. I'll do this today :)

tysonjhayes:
Bow to my experience?

You're making me out to be like this:
image

Instead of how I feel:
image

Feel free to tell me no. :laughing:

PlagueHO:
Hahaha LOL! :) I think I'm just mainly "borrowing" the best style elements and DSC Resource best practices based on what I'm seeing in the other DSC xResources - for better or for worse. So I'm definitely in this boat:
chemistry-dog
Still, I think these discussions are definitely the best ways of nutting out some good "best practices" - unless someone must wiser has already done this somewhere?

tysonjhayes:
Just wanted to follow up @PlagueHO did you have a chance to work on the changes?

PlagueHO:
Doh! I did, but I pushed them to my Dev branch, where as the PR is for my Rearchitect-xIPAddress-xDefaultGatewayAddress branch. Have merged them now. Sorry about that.

tysonjhayes:
@PlagueHO that merge only seems to have changes for the readme.md (unless I'm missing something). Also didn't pass the appvoeyor tests.

PlagueHO:
GitHub being helpful actually put the changes I made to remove the try/catches in the date/time order that I actually made them - so it slotted the commit message in quite a bit earlier. So the changes are in there, but you'll need to scroll way up to see the commit. You can see it with an ID version 61723c0 between the pictures. :) The Readme.md change was just to trigger an AppVeyor rebuild. The AppVeyor failure is a bit of a pain - it's the one where it fails to download the DSCTools module randomly and a rebuild just needs to usually be triggered. But I can only do that by checking something in. I'll do that now. Sorry about that!

tysonjhayes:
Yes, the failure is quite annoying I'm seeing it over in xWebadministration as well (https://github.com/PowerShell/xWebAdministration/issues/52) and am talking to people about trying to get it solved. It's clogging up quite a bit of work.

This looks way better, and way cleaner. I'll merge this once we get AppVeyor to pass the tests. :+1:

PlagueHO:
Doh! I logged the DSCResources.Test issue in with the DSCResource.Tests project before seeing your comments about talking to people. That will would be awesome to get that one resolved - I'm sure half my commits are "dummies" :)

tysonjhayes:
@PlagueHO one last thing, can we update the README.md to include these changes so we can get it communicated once we update the resource again?

PlagueHO:
Excellent point! And done! :+1:

tysonjhayes:
And Merged. Thanks for all of the hardwork.

PlagueHO:
A pleasure! I'll get onto issue #25 later on today :)

PR #31, opened 2015-10-16
xFirewall Changes to resolve Issue #25

PlagueHO:
This PR resolves issue #25 and allows built-in firewall rules to be enabled correctly.

It also contains additional changes:
* Message localization support added to xFirewall
* Foreach loops that handled when multiple Firewall rules were returned with a specified Name. This is not possible so the code was superfluous.
* DisplayGroup no longer used as an index in combination with Name when looking up a Firewall rule as it was not required.
* Readme.md updated with example of enabling a built-in firewall rule.
* Sample file Sample_xFirewall_EnableBuiltInFirewallRule.ps1 added.

### DIFF
--- a/Tests/Unit/MSFT_xFirewall.Tests.ps1
+++ b/Tests/Unit/MSFT_xFirewall.Tests.ps1
@@ -45,8 +45,6 @@ InModuleScope $DSCResourceName {
         Context 'Absent should return correctly' {
             Mock Get-NetFirewallRule
 
-                $breakvar = $true;
-
             It 'Should return absent' {
                 $result = Get-TargetResource -Name 'FirewallRule'
                 $result.Name | Should Be 'FirewallRule'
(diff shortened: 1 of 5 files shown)


Tests/Unit/MSFT_xFirewall.Tests.ps1
```diff
@@ -38,6 +38,9 @@ Import-Module -Name $DSCResourceModuleFile.FullName -Force
 $breakvar = $True
```
tysonjhayes: Would you do me the favor of removal all of these, I accidently checked them in, I use them to break into scripts and forgot to clean up after myself. :grimacing:

tysonjhayes:
Looks great! Now if we could only get AppVeyor to not be stupid so we don't have so many dummy commits.

Can I convince you to keep working on other DSC Resources? :grin: Kind of easy to make huge improvements when I actually have someone else to work with.

PlagueHO:
Amen to that :)

Tests/Unit/MSFT_xFirewall.Tests.ps1
PlagueHO: Absolutely! And done.

PlagueHO:
Doh! Sorry about that - I did a CTRL+F but I had an extra space on the beginning so only the 2nd occurance was found and I didn't use my eyes to see the other one :) Checked for breakvar this time in case of other whitespace issues. All done.

tysonjhayes:
Fantastic. Looks great!

PR #35, opened 2015-10-28
Minor Fix to xDNSServerAddress to correct verbose logging

PlagueHO:
This change does:
1. corrects verbose logging in xDNSServerAddress when mulitple DNS Server Addresses are passed.
2. Adds an ErrorAction Stop parameter to the Set-DnsClientServerAddress cmdlet to ensure it stops if validation DNS Server validation fails.
3. Assigns result of address comparison to variable to make it easier to read and understand the logic and changes the variable name to make it more obvious.
4. Ensure that result of Object-Compare always results in an Array.

### DIFF
--- a/Tests/Unit/MSFT_xDNSServerAddress.Tests.ps1
+++ b/Tests/Unit/MSFT_xDNSServerAddress.Tests.ps1
@@ -52,7 +52,7 @@ InModuleScope MSFT_xDNSServerAddress {
         }
         #endregion
 
-        Context 'comparing IPAddress' {
+        Context 'comparing IPv4 Address' {
             It 'should return true' {
 
                 $Splat = @{
@@ -64,11 +64,38 @@ InModuleScope MSFT_xDNSServerAddress {
                 $Result.IPAddress | Should Be $Splat.IPAddress
             }
         }
+
+        #region Mocks
+        Mock Get-DnsClientServerAddress -MockWith {
+
+            [PSCustomObject]@{
+                ServerAddresses = 'fe80:ab04:30F5:002b::1'
+                InterfaceAlias = 'Ethernet'
+                AddressFamily = 'IPv6'
+            }
+        }
+        #endregion
+
+        Context 'comparing IPv6 Address' {
+            It 'should return true' {
+
+                $Splat = @{
+                    Address = 'fe80:ab04:30F5:002b::1'
+                    InterfaceAlias = 'Ethernet'
+                    AddressFamily = 'IPv6'
+                }
+                $Result = Get-TargetResource @Splat
+                $Result.IPAddress | Should Be $Splat.IPAddress
+            }
+        }
+
     }
 
     #######################################################################################
 
     Describe 'Set-TargetResource' {
+
+        # Test IPv4
 
         #region Mocks
         Mock Get-DnsClientServerAddress -MockWith {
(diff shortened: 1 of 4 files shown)


tysonjhayes:
Change seems pretty straight forward, haven't looked at the tests in a while do you need to update anything there?

PlagueHO:
Good question. Actually looking at it, it might be a good idea to add a test that tests when multiple DNS addresses are supplied but only one exists. I'll do that. Other than that I can't think of anything else we could update with the tests (except test the verbose text output of the functions .... but that is a *slightly* larger task :) ).

PlagueHO:
Ok. I noticed there was a test missing for when no DNS Addresses exist but multiple were being added so I've added that. I also noticed that there were pretty much no IPv6 specific tests - and being seeing as IPv6 does have some slightly different behaviors to IPv4 I decided it might be prudent to add a set. So that is done too :)

olljanat:
Looks that this also fixes issue #36 
but DNS server IPs still not shown wrong on "DNSServersNotCorrectMessage".

Example on my lab verbose message is shown like this:
VERBOSE: [srv]:                            [[xDNSServerAddress]SetDNSsettings] Test-TargetResource: DNS Servers are not correct. Expected "System.String[]", actual "System.String[]".

PlagueHO:
Hi @olljanat,

Can you post the config for the xDNSServerAddresses resource you're using?

This is what I'm getting in the verbose logs using this latest update:

For primary only set:
```
VERBOSE: [SA_FS1]:                            [[xDNSServerAddress]IPv4D_2::[Networking]Network] Test-TargetResource: DNS Servers are not correct. Expected
"192.168.128.10", actual "".
```

For primary and secondary set:
```
VERBOSE: [SA_FS1]:                            [[xDNSServerAddress]IPv4D_2::[Networking]Network] Test-TargetResource: DNS Servers are not correct. Expected
"192.168.128.10,192.168.128.11", actual "".
```

And for IPv6:
```
VERBOSE: [SA_FS1]:                            [[xDNSServerAddress]IPv6D_2::[Networking]Network] Test-TargetResource: DNS Servers are not correct. Expected "fd53:ccc5:895a:bc00::a,fd53:ccc5:895a:bc00::b", actual "".
```

Thanks!

Tests/Unit/MSFT_xDNSServerAddress.Tests.ps1
```diff
@@ -64,12 +64,39 @@ InModuleScope MSFT_xDNSServerAddress {
...
+            [PSCustomObject]@{
+                ServerAddresses = 'fe80:ab04:30F5:002b::1'
+                InterfaceAlias = 'Ethernet'
+                AddressFamily = 'IPv6'
+            }
+        }
+        #endregion
+
+        Context 'comparing IPv6 Address' {
+            It 'should return true' {
```
tysonjhayes: Should probably update this to be more relavent to the case we're testing.

tysonjhayes:
Looks good to me. I've only got the one note and I'll get this merged right after you get that fixed. Sorry about the delay.

olljanat:
@PlagueHO, sorry, my mistake. I probably had old version of module on client earlier because now it works fine.

Tests/Unit/MSFT_xDNSServerAddress.Tests.ps1
PlagueHO: I added some more relevant tests to the Set/Test-TargetResource functions but I didn't to the get-targetresource because the ServerAddresses parameter is actually ignored by this function so I didn't think that adding the additional tests (with multiple addresses) would do anything. But I might be misunderstanding :)

tysonjhayes:
@PlagueHO is this good now? I'll merge if you're done.

PlagueHO:
@tysonjhayes: Yep! This is good to go. @olljanat : No worries! I've done this myself more than once. :)

PlagueHO:
:+1:

Issue #34, opened 2015-10-28
xFirewall: Fails on port keywords


nebffa:
The following DSC script:

```
Configuration WebServer
{
    Import-DSCResource -Module xNetworking

    Node "perftest01"
    {
        xFirewall "IIS Remote Administration"
        {
            Name = "IIS Remote Administration"
            DisplayName = "IIS Remote Administration"
            Ensure = "Present"
            Enabled = $true
            Action = "Allow"
            Profile = "Domain"
            Direction = "InBound"
            RemotePort = "Any"
            LocalPort = "PlayToDiscovery"
            Protocol = "TCP"
            Description = "Firewall rule for deployments to administer IIS."
        }
    }
}

WebServer
clear; Start-DscConfiguration -Force -Verbose -Wait -path .\WebServer
```

fails with the logs:

```
VERBOSE: Perform operation 'Invoke CimMethod' with following parameters, ''methodName' =
SendConfigurationApply,'className' = MSFT_DSCLocalConfigurationManager,'namespaceName' =
root/Microsoft/Windows/DesiredStateConfiguration'.
VERBOSE: An LCM method call arrived from computer PERFTEST01 with user sid
S-1-5-21-1778218145-2021600166-3147052739-5183.
VERBOSE: [PERFTEST01]: LCM:  [ Start  Set      ]
VERBOSE: [PERFTEST01]: LCM:  [ Start  Resource ]  [[xFirewall]IIS Remote Administration]
VERBOSE: [PERFTEST01]: LCM:  [ Start  Test     ]  [[xFirewall]IIS Remote Administration]
VERBOSE: [PERFTEST01]:                            [[xFirewall]IIS Remote Administration] Test-TargetResource: Checking
settings for firewall rule with Name 'IIS Remote Administration'.
VERBOSE: [PERFTEST01]:                            [[xFirewall]IIS Remote Administration] Test-TargetResource: Find
firewall rule with Name 'IIS Remote Administration'.
VERBOSE: [PERFTEST01]:                            [[xFirewall]IIS Remote Administration] Test-TargetResource: Check
each defined parameter against the existing Firewall Rule with Name 'IIS Remote Administration'.
VERBOSE: [PERFTEST01]:                            [[xFirewall]IIS Remote Administration] Get-FirewallRuleProperty: Get
all the properties and add filter info to rule map.
VERBOSE: [PERFTEST01]:                            [[xFirewall]IIS Remote Administration] Test-RuleProperties: LocalPort
[...]

PR #37, opened 2015-11-01
Issue34 and loging correction

PlagueHO:
This PR contains changes discussed in Issue #34. In adding the new pester tests to this resource, it also identified two minor issues (5 and 6) that I have resolved but may require further discussion.

The following things were changed:
1. Additional Pester tests added to check individual changes to the firewall rule properties.
2. Known Invalid Configurations section added to Readme.md.
3. Fix to Set-TargetResource so that if an existing rule is updated with a Known invalid configuration the correct error message will be displayed.
4. Fix to Set-TargetResource so that if DisplayName parameter is set when updating existing rule then error will not be thrown (actually same fix as 3).
5. Setting the ApplicationPath property would always cause an exception to be thrown, because no such parameter exists on the *-NetFirewallRule cmdlets. This parameter has been renamed to Program to match the cmdlets and to prevent the error. I felt that this was the best approach as this parameter could never have been used in the past anyway because it would **always** throw an exception - so I can't see any existing users of the resource are using this prop anyway.
6. Trying to change the DisplayGroup parameter (which actually changes the Group parameter) on an existing rule would throw an exception because it there is actually no way for the Set-NetFirewallRule cmdlet to change the group/display group property. This seems to me to be a limitation of the Cmdlet, so I've logged it on PowerShell connect. I've also noted this as a Known Invalid Config in the Readme.md. The issue is also detected and an appropriate exception is raised with advise to delete and re-create the rule instead. It would be possible to do this in the resource, but IMHO actually would cause more problems than it solves.

### DIFF
--- a/DSCResources/MSFT_xFirewall/MSFT_xFirewall.psm1
+++ b/DSCResources/MSFT_xFirewall/MSFT_xFirewall.psm1
@@ -416,7 +416,8 @@ function Test-RuleProperties
                 if (-not ($networkProfileinRule -contains $networkProfile))
                 {
                     Write-Verbose -Message ( @( "$($MyInvocation.MyCommand): "
-                        $($LocalizedData.PropertyNoMatchMessage) -f 'Profile',$networkProfileinRule,$Profile
+                        $($LocalizedData.PropertyNoMatchMessage) `
+                            -f 'Profile',($networkProfileinRule -join ','),($Profile -join ',')
                         ) -join '')
                     $desiredConfigurationMatch = $false
                     break
@@ -426,7 +427,8 @@ function Test-RuleProperties
         else
         {
             Write-Verbose -Message ( @( "$($MyInvocation.MyCommand): "
-                $($LocalizedData.PropertyNoMatchMessage) -f 'Profile',$networkProfileinRule,$Profile
+                $($LocalizedData.PropertyNoMatchMessage) `
+                    -f 'Profile',($networkProfileinRule -join ','),($Profile -join ',')
                 ) -join '')
             $desiredConfigurationMatch = $false
         }
@@ -435,7 +437,8 @@ function Test-RuleProperties
     if ($Direction -and ($FirewallRule.Direction -ne $Direction))
     {
         Write-Verbose -Message ( @( "$($MyInvocation.MyCommand): "
-            $($LocalizedData.PropertyNoMatchMessage) -f 'Direction',$FirewallRule.Direction,$Direction
+            $($LocalizedData.PropertyNoMatchMessage) `
+                -f 'Direction',$FirewallRule.Direction,$Direction
             ) -join '')
         $desiredConfigurationMatch = $false
     }
@@ -451,7 +454,8 @@ function Test-RuleProperties
                 if (-not ($remotePortInRule -contains($port)))
                 {
                     Write-Verbose -Message ( @( "$($MyInvocation.MyCommand): "
-                        $($LocalizedData.PropertyNoMatchMessage) -f 'RemotePort',$remotePortInRule,$RemotePort
+                        $($LocalizedData.PropertyNoMatchMessage) `
+                            -f 'RemotePort',($remotePortInRule -join ','),($RemotePort -join ',')
                         ) -join '')
                     $desiredConfigurationMatch = $false
                 }
@@ -460,7 +464,8 @@ function Test-RuleProperties
         else
         {
             Write-Verbose -Message ( @( "$($MyInvocation.MyCommand): "
-                $($LocalizedData.PropertyNoMatchMessage) -f 'RemotePort',$remotePortInRule,$RemotePort
+                $($LocalizedData.PropertyNoMatchMessage) `
+                    -f 'RemotePort',($remotePortInRule -join ','),($RemotePort -join ',')
                 ) -join '')
             $desiredConfigurationMatch = $false
         }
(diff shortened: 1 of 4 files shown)


Tests/Unit/MSFT_xFirewall.Tests.ps1
```diff
@@ -191,16 +191,176 @@ InModuleScope $DSCResourceName {
                 Assert-MockCalled Get-FirewallRule -Exactly 1
             }
         }
-        Context 'Ensure is Present and the Firewall Does Exist but is different' {
+        Context 'Ensure is Present and the Firewall Does Exist but has different DisplayName' {
```
tysonjhayes: Probably should be 'Has **a** different DisplayName'

DSCResources/MSFT_xFirewall/MSFT_xFirewall.psm1
```diff
@@ -188,8 +189,38 @@ function Set-TargetResource
                     $($LocalizedData.UpdatingExistingFirewallMessage) -f $Name
                     ) -join '')
 
-                # Set the existing Firewall rule based on specified parameters
-                Set-NetFirewallRule @PSBoundParameters
+                # If the DisplayGroup is being changed then the rule
+                # Has to be removed and recreated because that is the
+                # Only way the DisplayGroup can be changed unfortunately.
+                # A change to the set-netfirewallrule has been requested:
+                # https://connect.microsoft.com/PowerShell/feedbackdetail/view/1970765/add-ability-to-change-firewall-displaygroup-in-set-netfirewallrule-cmdlet
```
tysonjhayes: :+1: for actually filing a bug.

README.md
```diff
@@ -48,11 +48,21 @@ Please check out common DSC Resources [contributing guidelines](https://github.c
...
 * **Protocol**: Specific protocol for filter. Specified by name, number, or range.
 * **Description**: Documentation for the rule.
-* **ApplicationPath**: Path and filename of the program for which the rule is applied.
+* **Program**: Path and filename of the program for which the rule is applied.
 * **Service**: Specifies the short name of a Windows service to which the firewall rule applies.
 
+## Known Invalid Configurations
+
+### xFirewall
+* The exception 'One of the port keywords is invalid' will be thrown if a rule is created with the LocalPort set to PlayToDiscovery and the Protocol is not set to UDP. This is not an unexpected error, but because the New-NetFirewallRule documentation is ambiguous.
```
tysonjhayes: Would you be so kind as to file a connect for this too? We should request that the documentation be updated on technet for this as well.

README.md
```diff
@@ -48,11 +48,21 @@ Please check out common DSC Resources [contributing guidelines](https://github.c
...
+* The exception 'The DisplayGroup of an existing Firewall Rule can not be changed' will be thrown if a configuration tries to change DisplayGroup property of an existing rule. This is because the Set-NetFirewallRule cmdlet does not support this function. Delete and re-create this rule instead.
```
tysonjhayes: Could we add a note about the connect - https://connect.microsoft.com/PowerShell/feedbackdetail/view/1970765/add-ability-to-change-firewall-displaygroup-in-set-netfirewallrule-cmdlet - here as well? I'd like to draw more attention to it.

Tests/Unit/MSFT_xFirewall.Tests.ps1
```diff
@@ -191,16 +191,176 @@ InModuleScope $DSCResourceName {
...
+
+                Assert-MockCalled Set-NetFirewallRule -Exactly 1
+                Assert-MockCalled Test-RuleProperties -Exactly 1
+            }
+        }
+        Context 'Ensure is Present and the Firewall Does Exist but has different Profile' {
+            It "should call expected mocks on firewall rule $($rule.Name)" {
+                Mock Set-NetFirewallRule
+                Mock Test-RuleProperties {return $false}
+                if ($rule.Profile -ccontains 'Domain') {$NewProfile = @('Public','Private')} else {$NewProfile = @('Domain','Public')}
```
tysonjhayes: Can we put the { on a new line to get it under the character limit?  Also be cool if we put else on it's on line too..

My personal perference:

```powershell
if (test this)
{
   code
}
else
{
   else code
}

Not sure what the style guide is but let's at least break this up a little.

**Comment by tysonjhayes** [comment]
Everything looks good, annoying about the bugs you've found but awesome you've filed connects. Couple of quick things to address and I'll get this merged. Thanks so much!

File: README.md (hunk 3 above)
Comment (PlagueHO) [inline README.md]: Done!

https://connect.microsoft.com/PowerShell/feedbackdetail/view/1974268/new-set-netfirewallrule-cmdlet-localport-parameter-documentation-is-incorrect-for-playtodiscovery

I popped a note in the Community Additions on the Cmdlet pages as well.

File: Tests/Unit/MSFT_xFirewall.Tests.ps1 (hunk 5 above)
Comment (PlagueHO) [inline Tests/Unit/MSFT_xFirewall.Tests.ps1]: Good call :+1: I've made the changes and also done this to a few of the conditional statements in the Test-TargetResource pester test code. I also added a few more of these conditions to some of the other parameters to ensure that no matter which rule is selected to use as a test the test parameters will end up different.

The other alternative would be to use the PS equivalent of a ternary operator for these assignments, but some of them come in over the 100 char limit unfortunately.

e.g.
```powershell
$NewProfile = if ( $rule.Profile -ccontains 'Domain') { @('Public','Private') } else { @('Domain','Public') }
```

What I wouldn't give for a ( ? : ) style ternary operator in PS to thin the structure down. :+1:

PlagueHO:
All done! A pleasure as always. Cheers!

PlagueHO:
:+1:

Issue #41, opened 2015-11-06
Get-DscConfiguration Fails on Firewall Rule


PrahlM93:
Hello,
Using xNetworking 2.4, I am having issues running Get-DscConfiguration.

My configuration block looks as follows:
```
xFirewall RdpFirewallRules {
    Name = 'RemoteDesktop-UserMode-In-TCP'
    Ensure = 'Present'
    Action = 'Allow'
    Enabled = 'True'
    Profile = ('Domain', 'Private', 'Public')
}
```

I get the following error in Event Viewer when running Get-DscConfiguration:
```
This event indicates that failure happens when LCM is processing the configuration. Error Id is 0xD. Error Detail is The SendConfigurationApply function did not succeed.. Resource Id is [xFirewall]RdpFirewallRules and Source Info is C:\Configuration\Firewall.ps1::118::17::xFirewall. Error Message is The PowerShell DSC resource  returned results that are not valid from Get-TargetResource. The Group key is not a valid property in the corresponding DSC resource schema file. The results from Get-TargetResource must be in a Hashtable format. The keys in the Hashtable must be the same as the properties in the corresponding DSC resource schema file..
```

It seems that the Group property isn't defined in the schema. I ended up adding it as a read-only property and it seems to work but the information isn't particularly useful as shown below:
```
Group                : @FirewallAPI.dll,-28752
```

It seems that the appropriate solution is to have "Group" not be returned in the Get-TargetResource function. This is implemented in #42 

Thank you for your help.

tysonjhayes:
Howdy - I have a fix for this as well I was planning on adding some integration testing back into the code and managed to get it working with it and your suggested changes.

PR #42, opened 2015-11-06
Remove Group property from Get-TargetResource

PrahlM93:
This removes the Group property from the Get-TargetResource function because it isn't defined in the schema and it is causing Get-DscConfiguration to fail.

This is in reference to issue [#41](https://github.com/PowerShell/xNetworking/issues/41).

### DIFF
--- a/DSCResources/MSFT_xFirewall/MSFT_xFirewall.psm1
+++ b/DSCResources/MSFT_xFirewall/MSFT_xFirewall.psm1
@@ -71,7 +71,6 @@ function Get-TargetResource
         Name            = $Name
         Ensure          = 'Present'
         DisplayName     = $firewallRule.DisplayName
-        Group           = $firewallRule.Group
         DisplayGroup    = $firewallRule.DisplayGroup
         Enabled         = $firewallRule.Enabled
         Action          = $firewallRule.Action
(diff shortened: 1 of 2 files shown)


tysonjhayes:
Hi @PrahlM93 - good catch, I think you are correct that this is likely the source of the probem, I removed the key/value per your suggesiton and it does apepar to work.

However, I also see it in Test-ResourceProperty has $Group as well, could you please remove it from there also?

Thanks!

PrahlM93:
@tysonjhayes 
Thanks for the tip, I looked into it but it seems that removing it causes these lines to fail in Set-TargetResource:
```
# Effectively renaming DisplayGroup to Group
    if ($DisplayGroup) {
        $null = $PSBoundParameters.Add('Group', $DisplayGroup)
    }
    $null = $PSBoundParameters.Remove('DisplayGroup')
```

I'm not quite sure why DisplayGroup becomes Group, but in my opinion it might as well be kept as DisplayGroup since that is what the original parameter is called. I'd like your thoughts before further modifying it.

PlagueHO:
Hi @PrahlM93 ,

I think @tysonjhayes has included a fix for this in PR #43.

I added a note in PR #43 as to why this slightly odd treatment of Group/DisplayGroup occurs. I think it should be possible to rework the code slightly to include both a Group and DisplayGroup parameter - which is really what there should be. But I'll for feedback from @tysonjhayes to see what he thinks.

Thanks for working on this :+1:

Cosmo-Kramer:
@PlagueHO Thank you for pointing that out. You can go ahead and close/decline this PR since @tysonjhayes incorporated this fix in #43.

PR #43, opened 2015-11-07
Adding Firewall Integration testing

### LINKED ISSUE #41: Get-DscConfiguration Fails on Firewall Rule

tysonjhayes:
This readds Firewall Integration testing now tha tI've been able to figure out what was causing the previous failures. This also fixes #41 as I modified the changes to match the suggestions made by @PrahlM93 (whom I hope will review the change as well).

@PlagueHO - if you have some time to review this I'd appreciate it.


PlagueHO:
Good thinking on introducing Integration Testing - should definitely pick up things like issue #41! I'll also use your integration tests as a template and add them to other resources as well when a change is next required. Nicely done. :+1: 

I took a look and the code all looks 100% to me.

I know the Group/DisplayGroup stuff is because of the odd behavior of the ```new-netfirewallrule``` cmdlet, it doesn't support setting the DisplayGroup, even though ```set-netfirewallrule``` does. It would actually be possible to work around this limitation at some point (by firing a new-netfirewallrule then set-netfirewallrule when new rules are created that contain a DisplayGroup). That way both DisplayGroup and Group could be supported and set by the resource and more closely match what actually is stored internally. Right now it is a little bit strange. But I think that should be left for another time.

tysonjhayes:
@PlagueHO I think you are right we should fix up the DisplayGroup/Group. I've created #44 in order for us to get back to this. As I agree we should deal with this seperately.

As for treating this integration test as a template let me get a PR open to actually create a template so we can document the logic I used. Once we land on it we can try to get it put somewhere so the rest of the resources can benefit. :+1: 

Thanks for the review.

PlagueHO:
Good thinking! Might be nice to get other templates in there as well - for example, the header and footer of the unit/pester tests. I've tended to copy the ones you created and they work well in every scenario I've put them through.

Issue #44, opened 2015-11-09
xFirewall: Modify Set-TargetResource to support both Group/DisplayGroup


tysonjhayes:
Right now there is a limitation in the code where xFirewall doesn't support setting both Group and DisplayGroup, the code pretty much doesn't let us set Group. This should be changed to allow both.

PlagueHO:
:+1:

I could have a go at this change this week if you like? I've have some ideas on the changes that need to be made.

The fiddliest part is working around the issue where a DisplayGroup can't be renamed (https://connect.microsoft.com/PowerShell/feedbackdetail/view/1970765/add-ability-to-change-firewall-displaygroup-in-set-netfirewallrule-cmdlet) or where the Group needs to be renamed on an existing rule - because the only way to manage these two scenarios is to recreate the rule. But I think this can be managed.

PlagueHO:
Also, I added another request to Windows Connect to allow the Group to be changed in the Set-NetFirewallRule cmdlet:

https://connect.microsoft.com/PowerShell/feedbackdetail/view/2000226/add-ability-to-change-firewall-group-in-set-netfirewallrule-cmdlet

tysonjhayes:
One of these days we'll get everything open sourced so we can submit pull requests. :+1:

PlagueHO:
Now that would be awesome :) I'll happen one day I'm sure. Still, Connect is a pretty good way to start.

PlagueHO:
I've done some digging on this one and actually think it requires a slightly different solution that I originally thought. This does still require a change, so I'll summarize the situation here so that it is (hopefully) clear why this resource behaves this way and hopefully clarifies things.

My original thinking was there is no possible way to set the **DisplayGroup** on an existing or new Firewall rule. It will **always** be set to the value of the **Group** parameter that is passed in the ```New-NetFirewallRule``` cmdlet and once set can **not** be changed by any other cmdlet. It can't even be changed using NETSH or the WFAS UI!!

But after researching this I don't think the DisplayGroup is actually never meant to be set. We should ONLY ever allow the Group to be set. The DisplayGroup will always take on one of two values:
1. The same value as the Group.
2. If the group is starts with @ (this means it is a predefined rule) then the display group will be set from the DLL that is specified in the group. For example:
![2015-11-10_18-29-23](https://cloud.githubusercontent.com/assets/7589164/11055219/a048614a-87d9-11e5-9766-8cb23191eb18.png)
[...]

Issue #47, opened 2015-11-09
Missing "Scope" param for xFirewall


cydhaselton:
Cannot specify a set of Remote and or Local IP addresses/subnets for the xFirewall configuration

tysonjhayes:
Sorry I'm not that familiar with this configuration. Are you looking to specify RemoteMachine/RemoteAddress parameter for NetFirewalRule?


```powershell
-RemoteAddress<String[]>
Specifies that network packets with matching IP addresses match this rule. 
This parameter value is the second end point of an IPsec rule and specifies the computers that are subject to the requirements of this rule. 
This parameter value is an IPv4 or IPv6 address, hostname, subnet, range, or the following keyword: Any. 
The acceptable formats for this parameter are: 
-- Single IPv4 Address: 1.2.3.4 
-- Single IPv6 Address: fe80::1 
-- IPv4 Subnet (by network bit count): 1.2.3.4/24 
-- IPv6 Subnet (by network bit count): fe80::1/48 
-- IPv4 Subnet (by network mask): 1.2.3.4/255.255.255.0 
-- IPv4 Range: 1.2.3.4 through 1.2.3.7 
-- IPv6 Range: fe80::1 through fe80::9 
Note: Querying for rules with this parameter can only be performed using filter objects. See the Get-NetFirewallAddressFilter cmdlet for more information.

````

````powershell

-RemoteMachine<String>
Specifies that matching IPsec rules of the indicated computer accounts are created. 
This parameter specifies that only network packets that are authenticated as incoming from or outgoing to a computer identified in the list of computer accounts (SID) match this rule. This parameter value is specified as an SDDL string. 
Note: Querying for rules with this parameter can only be performed using filter objects. See the Get-NetFirewallSecurityFilter cmdlet for more information.

````

cydhaselton:
Correct.  I am looking to specify the remote address/machine parameter for NetFirewallRule within DSC.  
I was looking for the parameter in xFirewall...is it in a different location?

tysonjhayes:
I believe it should be in there, not sure why its not. It'll be a bit before I can spend some time looking at this and code up a solution.

I think it should be pretty straight forward. @PlagueHO might have time to get to this before I do but I'll try to get to it soon.

PlagueHO:
I'll take a look when I do PR #44 - should be fairly straight forward! :+1: I'll take a look at these tonight if you like?

tysonjhayes:
No rush on my account. Thanks!

[...]

PR #49, opened 2015-11-10
Fix for Issue #44 -xFirewall: Modify Set-TargetResource to support both Group/DisplayGroup 

PlagueHO:
This resolves issue #44.

It contains changes as per discussion in the Issue:
1. DisplayGroup renamed to Group and all code/documentation/samples/tests referencing it updated.
2. New DisplayGroup parameter added as a READ parameter so that it can be reported via Get-TargetResource.
3. If the Group value is set to a different value for an existing rule the rule will be deleted and recreated.
4. Integration Test code updated to automatically set execution policy (and restore it) so it will run on machines with a more restrictive policy.
5. Fix to ensure that if a resource should only update the DisplayName for an existing rule it will correctly detect that changes are required.

Hopefully this issue has been :hammer:

### DIFF
--- a/DSCResources/MSFT_xFirewall/MSFT_xFirewall.psm1
+++ b/DSCResources/MSFT_xFirewall/MSFT_xFirewall.psm1
@@ -183,7 +183,8 @@ function Set-TargetResource
 
                 # If the Group is being changed the the rule needs to be recreated
                 if ($PSBoundParameters.ContainsKey('Group') `
-                    -and ($Group -ne $FirewallRule.Group)) {
+                    -and ($Group -ne $FirewallRule.Group))
+                {
 
                     Remove-NetFirewallRule -Name $Name
 
@@ -192,17 +193,18 @@ function Set-TargetResource
 
                     # Loop through each possible property and if it is not passed as a parameter
                     # then set the PSBoundParameter property to the exiting rule value.
-                    'DisplayName','Group','Enabled','Action','Profile','Direction', `
-                    'RemotePort','LocalPort','Protocol','Description','Program','Service' `
-                        | Foreach-Object -Process {
-                            if (-not $PSBoundParameters.ContainsKey($_))
-                            {
-                                $PropertyValue = (Invoke-Expression -Command "`$FirewallRule.$_")
-                                if ($PropertyValue) {
-                                    $null = $PSBoundParameters.Add($_,$PropertyValue)
-                                }
+                    $PropList = @('DisplayName','Group','Enabled','Action','Profile','Direction', `
+                    'RemotePort','LocalPort','Protocol','Description','Program','Service')
+
+                    Foreach ($Prop in $PropsList) {
+                        if (-not $PSBoundParameters.ContainsKey($Prop))
+                        {
+                            $PropertyValue = (Invoke-Expression -Command "`$FirewallRule.$Prop")
+                            if ($PropertyValue) {
+                                $null = $PSBoundParameters.Add($Prop,$PropertyValue)
                             }
                         }
+                    }
 
                     New-NetFirewallRule @PSBoundParameters
                 }
--- a/DSCResources/MSFT_xFirewall/MSFT_xFirewall.psm1
+++ b/DSCResources/MSFT_xFirewall/MSFT_xFirewall.psm1
@@ -198,7 +198,9 @@ function Set-TargetResource
                             if (-not $PSBoundParameters.ContainsKey($_))
                             {
                                 $PropertyValue = (Invoke-Expression -Command "`$FirewallRule.$_")
-                                $null = $PSBoundParameters.Add($_,$PropertyValue)
+                                if ($PropertyValue) {
+                                    $null = $PSBoundParameters.Add($_,$PropertyValue)
+                                }
                             }
                         }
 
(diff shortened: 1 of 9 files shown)


DSCResources/MSFT_xFirewall/MSFT_xFirewall.psm1
```diff
@@ -178,34 +177,34 @@ function Set-TargetResource
...
                     ) -join '')
 
-                # If the DisplayGroup is being changed then the rule
-                # Has to be removed and recreated because that is the
-                # Only way the DisplayGroup can be changed unfortunately.
-                # A change to the set-netfirewallrule has been requested:
-                # https://connect.microsoft.com/PowerShell/feedbackdetail/view/1970765/add-ability-to-change-firewall-displaygroup-in-set-netfirewallrule-cmdlet
+                # If the Group is being changed the the rule needs to be recreated
                 if ($PSBoundParameters.ContainsKey('Group') `
                     -and ($Group -ne $FirewallRule.Group)) {
```
tysonjhayes: nit: can we move the { to a new line to match the style elsewhere?

README.md
```diff
@@ -356,14 +356,14 @@ Configuration Add_FirewallRuleToExistingGroup
         {
             Name                  = "MyFirewallRule"
             DisplayName           = "My Firewall Rule"
-            DisplayGroup          = "My Firewall Rule Group"
+            Group          = "My Firewall Rule Group"
```
tysonjhayes: nit: can you get the spacing to line up on this line? Always drives me nuts when it doesn't. :grin:

Tests/Integration/MSFT_xFirewall.Integration.Tests.ps1
```diff
@@ -30,9 +30,14 @@ if (($env:PSModulePath).Split(';') -ccontains $pwd.Path)
     $env:PSModulePath = ($env:PSModulePath -split ';' | Where-Object {$_ -ne $pwd.path}) -join ';'
 }
 
+# Preserve and set the execution policy so that the DSC MOF can be created
+$OldExecutionPolicy = Get-ExecutionPolicy
```
tysonjhayes: I'd suggest here instead doing the following.

````powershell
$executionPolicy = Get-ExecutionPolicy
if ($executionPolicy -ne 'Unrestricted')
{
    Set-ExecutionPolicy -ExecutionPolicy Unrestricted -Force
    $rollbackExecution = $true
}

#### do everything here ###

Finally {
   if ($rollbackExection)
   {
        Set-ExectuionPolicy $executionPolicy
    }
}
````

This way we're not attempting to change an Execution Policy that doesn't need to be changed.

File: DSCResources/MSFT_xFirewall/MSFT_xFirewall.psm1 [hunk 4]
```diff
@@ -178,34 +177,34 @@ function Set-TargetResource
...
-                    $PSCmdlet.ThrowTerminatingError($errorRecord)
+
+                    Remove-NetFirewallRule -Name $Name
+
+                    # Merge the existing rule values into the PSBoundParameters
+                    # so that it can be splatted.
+
+                    # Loop through each possible property and if it is not passed as a parameter
+                    # then set the PSBoundParameter property to the exiting rule value.
+                    'DisplayName','Group','Enabled','Action','Profile','Direction', `
```
tysonjhayes: Couple of suggestions here.

1) Piping is actually slower then foreach, so I'd suggest changing this to foreach ($parameter in $list) which wills peed it up.

2) Move this to it's own array as a seperate variable so we can modify this as needed. Though it would be aweosme to autodiscover in some way. In theory we could run $parameters = (Get-Command Get-NetFirewallRule).parameters but you'll get way more stuff then you're intending to get.

PlagueHO: Good idea! Done. I did look into auto pulling the properties list (it would need to get pulled from new-netfirewallrule), but because at the moment we're missing about 50% of the properties it didn't make sense. But once we've implemented all of them it will (although we'll still need to strip out ones implemented by the PSCmdLet base class and a few others).

tysonjhayes: [System.Management.Automation.PSCmdlet]::CommonParameters
[System.Management.Automation.PSCmdlet]::OptionalCommonParameters

Those two properties will show all of the common/optional parameters so we can filter them out. I was playing with it for a minute but got pulled away.

PlagueHO: Although we'd want to use:
```powershell
(Get-Command New-NetFirewallRule).parameters
```
as this is the parameter set that we'll be using to splat onto the New. But as noted in Issue #47 , we're definitely missing a few :(

tysonjhayes: Er. Right, that's what I meant. :grinning:  Yes, let's get that fixed up, when we get this submitted let's get an issue open to fix it once we've added support for the other parameters (or do it as part of the support I guess :+1: )

Tests/Integration/MSFT_xFirewall.Integration.Tests.ps1
```diff
@@ -62,6 +72,12 @@ try {
     }
 }
 finally {
+    # Restore the Execution Policy
+    if ($rollbackExection)
```
tysonjhayes: Should be $rollbackExecution

tysonjhayes:
Ok once the one line is fixed I'll get this merged. Thanks for the hardwork!

PlagueHO:
Done and done! :+1:

tysonjhayes:
Well at least we caught that mistype too. :+1: one less bug!

PlagueHO:
Except I missed the other one! Oh dear. Fixed now! :)

PR #51, opened 2015-11-11
Adding NetConnectionProfile DSC Resource

### LINKED ISSUE #19: Create NetConnectionProfile resource

tysonjhayes:
This commit allows the modification of Get/Set-NetConnectionProfile in DSC.
It also adds unit testing for the resource. We'll want to add Integration testing as a seperate commit to allow further refinement on this resource to occur.

Resolves #19 as this allows the modification of this command.


tysonjhayes:
@Dan1el42 @iainbrighton @PlagueHO - if anyone of you have time for a code review on this i'd be appreciated. 

I'd like to get in some intergration tests before I commit but I'd like some passes on the logic before I start figuring out that process.

PlagueHO:
The logic looks good to me. No problems I can see. :+1:

Tests/Unit/MSFT_xFirewall.Tests.ps1
```diff
@@ -149,6 +149,7 @@ try
                 }
             }
             Context 'Ensure is Present and the Firewall is Absent' {
+                Mock Get-FirewallRule
```
tysonjhayes: @PlagueHO - I had to add this to get the unit tests to pass on AppVoyer (no idea why it was working previously) can you validate the logic is still sound with this change and we're testing the right things?

PlagueHO: Ok, I've figured it out.

Without the Mock the Test-FirewallProperty would fire because the rule existed. Although in this test we're trying to ensure the function returns $true (there is a difference). But some rules will report no difference if the rule exists because all values in the rule are just the default. So whether this test passed or not (without the new mock) would be down to the rule tested. Introducing the Mock corrected the problem.

What I did to figure this out was wrap the entire unit test code block in a foreach that tested every single rule on my computer (we could do this for real if we wanted, but it will take hours to run- if not days on AppVeyor). Without your the new Mock some rules would fail which allowed me to track and identify it what was going on.

So you definitely identified and fixed a nasty bug in the Unit tests there. The actual logic in the resource is OK.

PlagueHO: One more quick word about this:
As a test I thought I'd let this run through every rule on my Win 10 desktop (with your above fix) thinking that there would be no failures. Unfortunately there were 2 failed tests out of 45693 tests:


I didn't output the nunit XML here though so I can't find the failed tests until I get a chance to rerun them (which takes about 4 hours). I'll do this tomorrow and track down what ever bug still remains. I'll log this in a separate issue/PR when I find it.

tysonjhayes: I'd be interested in seeing the code you used to test everything it might be interesting if we can get it to go fast enough to do a pass on everything (occassionally).

tysonjhayes: Ok I'll play around with it let you know if I figured out something nice, I think I see some room for improvment but I'll have to play around with it a bit.

PR #46, opened 2015-11-09
type(test): Adding Integration Template for future Integration tests

tysonjhayes:
This adds a template for how we expect future integration tests to look.
It allows us to show best practices for backwards compatability and allows
 us to formalize the template.

ISSUE_KEY: https://github.com/PowerShell/xNetworking/issues/45

PR #52, opened 2015-11-12
Issue 47 - Added Missing Scope Parameters

PlagueHO:
This PR addresses the Issue #47 Missing "Scope" Parameters as well as additional work on #46.

It contains the following changes:
1. xFirewall Parameters added:
    * Authentication
    * Encryption
    * InterfaceAlias
    * InterfaceType
    * LocalAddress
    * LocalUser
    * Package
    * Platform
    * RemoteAddress
    * RemoteMachine
    * RemoteUser
2. xFirewall resource reworked internally to use an Array that contains a list of all parameters and the source data elements and this is used in both the resource (Get/Set/Test) and the pester tests to eliminate code replication. The array is defined in data region called ParameterList in MSFT_xFirewall.psm1.
3. The profile parameter is now handled correctly as an array internally instead of being converted back and forth between comma delimited string and array.
4. Array comparison for array parameters now done using Compare-Object instead of loop and count comparison.
5. Unit Test template added.
6. Unit Test and Integration Test header/footer code standardized and copied to templates. Header/Footer code should be the same in all tests and test templates. Header/Footer code can now be moved into external helper module like DSCResource.Tools.
7. Integration Test configuration file renamed to MSFT_xFirewall.config.ps1 so that the Integration Test template can automatically figure out the expected name without the user having to set it.
8. Also changed ApplicationPath to Program in samples and Readme.md as it was changed a few PRs ago.

That is it for this PR I think.

There are still a few parameters needing to be added, but they should be much easier to add now that the parameter list is in an array. I'll add the remaining parameters to a new Issue so we can track that.

Thanks!


DSCResources/MSFT_xFirewall/MSFT_xFirewall.psm1
```diff
@@ -57,31 +96,47 @@ function Get-TargetResource
...
-        Protocol        = $properties.PortFilters.Protocol
-        Program         = $properties.ApplicationFilters.Program
-        Service         = $properties.ServiceFilters.Service
+    $Result = @{
+        Ensure = 'Present'
+    }
+
+    # Populate the properties for get target resource by looping through
+    # the parameter array list and adding the values to 
+    foreach ($p in $ParameterList)
```
tysonjhayes: Could we call this foreach ($parameter in $parameterlist) so it's clearer what we're iterating on?

DSCResources/MSFT_xFirewall/MSFT_xFirewall.psm1
```diff
@@ -57,31 +96,47 @@ function Get-TargetResource
...
+    {
+
+        if ($p.type -eq 'Array')
```
tysonjhayes: Is the real difference in code here the verbose message?

If so I'm thinking one of two things.

1) Cut it, I mean it's nice to know don't get me wrong but seems strange to fork the code here just for a better formatted verbose message.
2) Move the $Value and $Result out of the if/else bracket and just have it run the different verbose message. This reduces the duplicated code and maintains the verbose message.  Also clarifies where the fork in the logic is actually occuring.

DSCResources/MSFT_xFirewall/MSFT_xFirewall.psm1
```diff
@@ -408,181 +552,62 @@ function Test-RuleProperties
...
-                if (-not ($remotePortInRule -contains($port)))
-                {
-                    Write-Verbose -Message ( @( "$($MyInvocation.MyCommand): "
-                        $($LocalizedData.PropertyNoMatchMessage) `
-                            -f 'RemotePort',($remotePortInRule -join ','),($RemotePort -join ',')
-                        ) -join '')
-                    $desiredConfigurationMatch = $false
+            'Array' {
+                # Array comparison uses Compare-Object
+                if ($ParameterSource -eq $null) {
```
tysonjhayes: Curly Braces on new lines please.

DSCResources/MSFT_xFirewall/MSFT_xFirewall.psm1
```diff
@@ -408,181 +552,62 @@ function Test-RuleProperties
...
+            'String' {
+                # Perform a plain string comparison.
+                if ($ParameterNew -and ($ParameterSource -ne $ParameterNew))
                 {
                     Write-Verbose -Message ( @( "$($MyInvocation.MyCommand): "
                         $($LocalizedData.PropertyNoMatchMessage) `
-                            -f 'Profile',($networkProfileinRule -join ','),($Profile -join ',')
+                            -f $p.Name,$ParameterSource,$ParameterNew
                         ) -join '')
                     $desiredConfigurationMatch = $false
```
tysonjhayes: Given at this point we have our tests down to two places where we can hit a false would it just make sense to just return $false here instead of assigning the variable? It means we should exit out of the code sooner instead of continuing to iterate through.

Also means the logic comes down to
````powershell
'String'
{
   # Parameter doesn't match
return $false
}
'Array'
{
   # do array compare
  return $false
}
````

File: DSCResources/MSFT_xFirewall/MSFT_xFirewall.psm1 [hunk 5]
```diff
@@ -408,181 +552,62 @@ function Test-RuleProperties
...
-
-        if ($networkProfileinRule.Count -eq $Profile.Count)
+        $ParameterSource = (Invoke-Expression -Command "`$($($p.source))")
+        $ParameterNew = (Invoke-Expression -Command "`$$($p.name)")
+        switch ($p.type)
         {
-            foreach($networkProfile in $Profile)
-            {
-                if (-not ($networkProfileinRule -contains $networkProfile))
+            'String' {
```
tysonjhayes: Curly brace on a new line please.

Templates/integration_template.ps1
```diff
@@ -41,50 +73,75 @@ if (($env:PSModulePath).Split(';') -ccontains $pwd.Path)
...
             } | Should not throw
         }
 
         It 'should be able to call Get-DscConfiguration without throwing' {
-            {Get-DscConfiguration} | Should Not throw
+            { Get-DscConfiguration -Verbose -ErrorAction Stop } | Should Not throw
         }
+        #endregion
 
         It 'Should have set the firewall and all the parameters should match' {
```
tysonjhayes: This should probably say 'Should have set the resource and all the parameters should match' to be more generic.

Templates/integration_template.ps1
```diff
@@ -41,50 +73,75 @@ if (($env:PSModulePath).Split(';') -ccontains $pwd.Path)
...
-            # Vaidate the config was set correctly
+            # TODO: Validate the Config was Set Correctly Here...
         }
 
     }
 }
+
+####################################################################################################
```
tysonjhayes: To be honest I don't love this as a formatting style, though I'm not really sure I have any good suggestions on what I don't like about it.

Are you just trying to visually break up the tests?

Tests/Integration/MSFT_xFirewall.Integration.Tests.ps1
```diff
@@ -37,50 +70,82 @@ if ($executionPolicy -ne 'Unrestricted')
...
-    # Load in the DSC Configuration
-    . $PSScriptRoot\Firewall.ps1
+####################################################################################################
 
-    Describe 'xFirewall_Integration' {
+    <#
+      This file exists so we can load the test file without necessarily having xNetworking in
+      the $env:PSModulePath. Otherwise PowerShell will throw an error when reading the Pester File.
+    #>
+    $ConfigFile = Join-Path -Path $PSScriptRoot -ChildPath "$DSCResourceName.Config.ps1"
```
tysonjhayes: Should probably be a lowercase C to match the file name: $DSCResourceName.config.ps1

Tests/Integration/MSFT_xFirewall.Integration.Tests.ps1
```diff
@@ -37,50 +70,82 @@ if ($executionPolicy -ne 'Unrestricted')
...
-            $firewallRule.Name         | Should Be $rule.Name
-            $firewallRule.DisplayName  | Should Be $rule.DisplayName
-            $firewallRule.Group        | Should Be $rule.Group
-            $firewallRule.Enabled      | Should Be $rule.Enabled
-            $firewallRule.Profile      | Should Be $rule.Profile
-            $firewallRule.Action       | Should Be $rule.Action
-            $firewallRule.Description  | Should Be $rule.Description
-            $firewallRule.Direction    | Should Be $rule.Direction
+            # Use the Parameters List to perform these tests
+            foreach ($p in $ParameterList)
```
tysonjhayes: Could we do foreach ($parameter in $parameterlist) please?

Tests/Integration/MSFT_xFirewall.Integration.Tests.ps1
tysonjhayes: Also good call on renaming the file. Much easier to understand.

Templates/unit_template.ps1
```diff
@@ -0,0 +1,160 @@
...
+}
+#endregion
+
+# TODO: Other Optional Init Code Goes Here...
+
+# Begin Testing
+try
+{
+
+####################################################################################################
```
tysonjhayes: To be honest I don't love this as a formatting style, though I'm not really sure I have any good suggestions on what I don't like about it.

Are you just trying to visually break up the tests?

if so could we just use ````#region <function name>```` instead of a giant block of # symbols?

DSCResources/MSFT_xFirewall/MSFT_xFirewall.psm1
```diff
@@ -191,17 +291,16 @@ function Set-TargetResource
...
+                    $properties = Get-FirewallRuleProperty -FirewallRule $firewallRule
+
                     # Loop through each possible property and if it is not passed as a parameter
                     # then set the PSBoundParameter property to the exiting rule value.
-                    $PropList = @('DisplayName','Group','Enabled','Action','Profile','Direction', `
-                    'RemotePort','LocalPort','Protocol','Description','Program','Service')
-
-                    Foreach ($Prop in $PropsList) {
-                        if (-not $PSBoundParameters.ContainsKey($Prop))
+                    Foreach ($p in $ParametersList) {
```
tysonjhayes: Could we make this $parameter in $parameterslist please?

DSCResources/MSFT_xFirewall/MSFT_xFirewall.psm1
```diff
@@ -408,181 +552,62 @@ function Test-RuleProperties
...
+        [String] $RemoteMachine,
+        [String] $RemoteUser
     )
 
     $properties = Get-FirewallRuleProperty -FirewallRule $FirewallRule
 
     $desiredConfigurationMatch = $true
 
-    if ($Name -and ($FirewallRule.Name -ne $Name))
+    # Loop throug the ParameterList array and compare the source
```
tysonjhayes: Loop through

DSCResources/MSFT_xFirewall/MSFT_xFirewall.psm1
```diff
@@ -408,181 +552,62 @@ function Test-RuleProperties
...
+    # with the value of each parameter. If different then
+    # set $desiredConfigurationMatch to false.
+    foreach ($p in $ParameterList)
```
tysonjhayes: $parameter in $parameterlist please.

tysonjhayes:
Looks pretty good, most of my stuff is minor. I think some of the style stuff chosen needs to be hashed out in a broader sense so I might end up merging it and then bringing up the point else where.

DSCResources/MSFT_xFirewall/MSFT_xFirewall.psm1
PlagueHO: Sure can! Consider it done.

DSCResources/MSFT_xFirewall/MSFT_xFirewall.psm1
PlagueHO: There is one tiny tiny difference between the two - but it is an important one (and super easy to over look):
```powershell
$Value = @(Invoke-Expression -Command "`$($($parameter.source))")
```
vs.
```powershell
$Value = (Invoke-Expression -Command "`$($($parameter.source))")
```
For the array (the first one) I have to explicitly cast this to an Array because otherwise if there is only a single array element PS will cast it to the type that the item is rather than keep it as an array.

So originally I just had it the if/then logic and no logging at all but I was trying to track down an issue occurring with the Get-DSCConfiguration logic so added that verbose logging. It really isn't needed anymore. So if you want I'm happy to get rid of the logs (but leave the if/then).

DSCResources/MSFT_xFirewall/MSFT_xFirewall.psm1
PlagueHO: Doh! I've been trying to make sure I do this but must have missed these. All fixed.

DSCResources/MSFT_xFirewall/MSFT_xFirewall.psm1
PlagueHO: That's a really good point. I guess it depends if we want to iterate over each test even it fails to write in the log each parameter that mismatches. For debugging of resources this might be useful, but I doubt it is critical. I could go either way with this one. Your call :)

DSCResources/MSFT_xFirewall/MSFT_xFirewall.psm1
PlagueHO: :+1:

Templates/integration_template.ps1
PlagueHO: Good call. Done.

Tests/Integration/MSFT_xFirewall.Integration.Tests.ps1
PlagueHO: Done!

Tests/Integration/MSFT_xFirewall.Integration.Tests.ps1
PlagueHO: :+1: What would we do without refactoring tools? :smile:

Templates/unit_template.ps1
PlagueHO: I think I originally copied this style out of some resource or PSM1 somewhere along the way. But I'm completely OK with whatever. I do like to break up the sections visually, but I'm good with anything you suggest really - I don't have a specific preference.

So I'll go with the Regions. Consider it done.

DSCResources/MSFT_xFirewall/MSFT_xFirewall.psm1
PlagueHO: Refactored :+1:

DSCResources/MSFT_xFirewall/MSFT_xFirewall.psm1
PlagueHO: Good eye! Wish ISE had spell check in comments like VS.

DSCResources/MSFT_xFirewall/MSFT_xFirewall.psm1
PlagueHO: All done!

PlagueHO:
Once this has been merged I'll adjust all the other resources in this module with the changes as well as add integration tests. I'll also remove all the ################'s when I do this.

Tests/Integration/MSFT_xFirewall.Integration.Tests.ps1
tysonjhayes: Unfortunately looks like you missed this one. But refactoring tools are amazing.

PlagueHO: Sorry about that! Done. As a matter of interest, what tools do you use for PS dev? I originally used VS 2013 wish POSH tools (and still do for some projects), but now I find myself using ISE (with ISE Steroids). I'd love it if VS Code had PS support though simply because how fast and light it is. :)

PR #58, opened 2015-11-19
Fix Firewall Unit Tests and Allow Multiple Firewall rules to be tested

PlagueHO:
This PR changes:
1. Fixes xFirewall Unit tests to prevent false positive failures caused by the values that firewall properties are changed to actually matching the existing values. This was detected only when I tested ALL firewall rules available on my computer as noted in PR #51 .
2. Added a foreach loop around the Firewall pester tests so that more than one rule could be tested - note: It still only currently tests the first rule, so the actual behavior hasn't changed here.


tysonjhayes:
The diff isn't showing for some reason so this is going to be kind of hard to review. I'll have to pull the change in locally and review it on a diff tool on my machine.

However, my first thought is number 2 should be removed, it's just confusing why we are doing a foreach loop on just one result and a waste of cycles. I'll spend some time looking at doing a loop over everything and we can add the functionality then.

PlagueHO:
Oh. That is strange!? I can recreate the PR easily enough if you like - save you the trouble?

The only reason I left the loop in was that I thought we could change the logic to pull the rules to test (in future if you're able to figure out a way to optimize the code) to pull multiple rules - all of them if possible. But if the code can't be optimized, I thought we could either pull say the first 10 and test those OR perhaps pull a random set of 10 or so out (using  a modulus function to choose them or something like that). Ideally I think testing more than one rule might be good. Will totally take your lead on this :)

tysonjhayes:
Lets not try to future proof now, if we think a random sample of 10 is needed, let's put that in. If we want to wait until we can see how many we can reasonably test at once then let's handle that with a seperate PR and remove the for each loop for now.

PlagueHO:
:+1: You are correct - and Agile development practices complete agree with you (I came from a waterfall world). Would you mind if I close and recreate this PR? I am not overly happy with the fact that the Diff hasn't worked.

tysonjhayes:
It's your PR, feel free to close it. :+1:

PlagueHO:
Thanks. Will raise a new one momentarily. Cheers!

PR #59, opened 2015-11-19
Firewall unit test fix

PlagueHO:
Fixes:
1. Corrects Test-TargetResource and Set-TargetResource tests to ensure that when a parameter is being tested with a different value the value is **actually** different.
2. Ensured that ALL tests now use the same Firewall rule to test with (rather than change halfway through the tests).
3. Added Resource name to the Describe cmdlet so that what is being tested is more easily identifiable (also, makes NUNIT output useful).
4. Added firewall rule name being tested to each 'It' command so that NUINT output is usable - also makes them consistent.


tysonjhayes:
I some how missed this. My apologizes. This looks great.

PlagueHO:
No worries! Cheers sir! :+1:

PR #62, opened 2015-12-12
Fix $env:PSModule getting set to null

tysonjhayes:
In the affected code $env:PSModulePath was being set to null when $script:tempPath did not contain a value.

/cc: @PlagueHO

### DIFF
--- a/Templates/unit_template.ps1
+++ b/Templates/unit_template.ps1
@@ -128,7 +128,10 @@ finally
 {
     #region FOOTER
     # Set PSModulePath back to previous settings
-    $env:PSModulePath = $script:tempPath;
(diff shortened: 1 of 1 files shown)


PlagueHO:
Good catch!

I made a similar change in PR https://github.com/PowerShell/DscResource.Tests/pull/11 (the one to add the Test Environment create/tear down code in these templates TestHelpers.psm1).

https://github.com/PlagueHO/DscResource.Tests/blob/master/TestHelper.psm1#L389.

tysonjhayes:
OK cool, just wanted to make sure you've adjusted for it else where. :+1:

Issue #32, opened 2015-10-17
xIPAddress: Incorrectly reports that DHCP is enabled for IPv6 addresses


PlagueHO:
The xIPAddress resource when used to set a static IPv6 address will misreport that the interface on the adapter is configured with DHCP when it is not. This causes Test-TargetResource to return $False, triggering the IP Address to be updated every time the config is applied.

The problem appears to be in the Get-NetIPInterface cmdlet at line 234 of MSFT_xIPAddress.psm1. I am referring to the current Dev version of this resource. This cmdlet always returns that DHCP is enabled even when it is not:

![pic](https://cloud.githubusercontent.com/assets/7589164/10557287/30bbbb90-7502-11e5-9e6d-b678c8d12f4b.png)

Note, the IPv4 address in the above screenshot is also statically assigned. Yet, the Get-NetIPInterface cmdlet returns that DHCP is enabled on the IPv6 interface but disabled on the IPv4.

Is this a problem in the cmdlet itself or does the cmdlet behave differently with IPv6? I know that IPv6 is a little more complex than IPv4 with things like Link-Local addresses and Site-Local/Unique-Local being assigned to the same interface, so perhaps this is a normal behavior?

The problem code is this:
```powershell
if (-not (Get-NetIPInterface `
        -InterfaceAlias $InterfaceAlias `
        -AddressFamily $AddressFamily).Dhcp.ToString().Equals('Disabled'))
```

To fix this, it needs to be changed to something that reliably detects the DHCP state of an IPv6 interface. Possibly the using the suffixOrigin or PrefixOrigin properties returned on objects from the Get-NetIPAddress cmdlet - if they equal 'manual' then DHCP is disabled. But this feels a bit risky to me. 

Any ideas?

PlagueHO:
And another example (except in a Windows Server 2012 R2 Full install):
![pic2](https://cloud.githubusercontent.com/assets/7589164/10561792/a463317e-7596-11e5-9809-a996f3055262.png)

PlagueHO:
After further review I think the problem can be resolved by manually setting the DHCP status on the IPv6 Interface in Set-TargetResource when the IP address is set. Alternately the check to ensure DHCP is set could be removed completely. This seems to make more sense when taking Issue #17 into account (moving DHCP into a separate resource).

My feeling is that the check for the DHCP setting should be removed completely.

PlagueHO:
[...]

Issue #17, opened 2015-09-02
Breakup IP Address Resource to DHCP and Static


tysonjhayes:
I think xIPAddress is trying to cover the gambit on what is really two distinct things.  One use case would be trying to assign a NIC a static IP address.  The other would be to assign the NIC to a DHCP server.

As the NIC can not have a static IP and be assigned to a DHCP server I prepose we break up the resources to handle those two different tasks.  This way we could have StaticIPAddress and DHCPAddress resources that just handle those use cases.

Thoughts?

Dan1el42:
What about using Ensure = Absent to enable DHCP on a network interface instead breaking up the resource? If you specify Ensure = Present the IPAddress property must be required to configure an IP address.

Wouldn't that provide the same consistency and follow more closely the DSC concepts?

"Just my 2 cents"

BladeFireLight:
That only makes sense if the Absent IP is the only IP on the NIC. Why not have the IP address take a string and allow DHCP as the value. then all IP's can be removed and DHCP turned on.

tysonjhayes:
I guess the way I'm thinking of it is the division is really around how we implement the logc.

To borrow from [a blog post](http://www.adminarsenal.com/admin-arsenal-blog/using-powershell-to-set-static-and-dhcp-ip-addresses-part-1) I found setting up a static IP is done this way.

```powershell
$IP = "10.10.10.10"
$MaskBits = 24 # This means subnet mask = 255.255.255.0
$Gateway = "10.10.10.1"
$Dns = "10.10.10.100"
$IPType = "IPv4"

# Retrieve the network adapter that you want to configure
$adapter = Get-NetAdapter | ? {$_.Status -eq "up"}

# Remove any existing IP, gateway from our ipv4 adapter
If (($adapter | Get-NetIPConfiguration).IPv4Address.IPAddress) {
    $adapter | Remove-NetIPAddress -AddressFamily $IPType -Confirm:$false
}

If (($adapter | Get-NetIPConfiguration).Ipv4DefaultGateway) {
    $adapter | Remove-NetRoute -AddressFamily $IPType -Confirm:$false
}

 # Configure the IP address and default gateway
$adapter | New-NetIPAddress `
    -AddressFamily $IPType `
    -IPAddress $IP `
    -PrefixLength $MaskBits `
    -DefaultGateway $Gateway

# Configure the DNS client server IP addresses
$adapter | Set-DnsClientServerAddress -ServerAddresses $DNS
````

[...]

PR #67, opened 2015-12-20
Fix for Issue #32 - Changed xIPAddress so that DHCP status is not checked for IPv6

PlagueHO:
This is a fix for #32.

The change prevents the test for DHCP status to be performed on IPv6 addresses. If this test is allowed then Test-TargetResource will return $false even if the IPv6 address is correctly assigned and set to static. This causes the resource to continuously apply the IPv6 address every time it runs - which can cause intermittent network connectivity issues.

This change actually only wraps the DHCP test code in a check to see if it is IPv4.

### DIFF
--- a/DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1
+++ b/DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1
@@ -17,6 +17,7 @@ SubnetMaskDoesNotMatchMessage=Subnet mask does NOT match desired state. Expected
 SubnetMaskMatchMessage=Subnet mask is in desired state.
 DHCPIsNotDisabledMessage=DHCP is NOT disabled.
 DHCPIsAlreadyDisabledMessage=DHCP is already disabled.
+DHCPIsNotTestedMessage=DHCP status is ignored when Address Family is IPv6.
 InterfaceNotAvailableError=Interface "{0}" is not available. Please select a valid interface and try again.
 AddressFormatError=Address "{0}" is not in the correct format. Please correct the Address parameter in the configuration and try again.
 AddressIPv4MismatchError=Address "{0}" is in IPv4 format, which does not match server address family {1}. Please correct either of them in the configuration and try again.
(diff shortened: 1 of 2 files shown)


tysonjhayes:
Could we add a warning or something to indicate to the operator that DHCP for IPv6 isn't supported?

tysonjhayes:
Like
````powershell
else {
    Write-Warning 'DHCP  for IPv6 is not supported in this module'
}

PlagueHO:
I reckon you're right - we should add some sort of message. But need to figure out the wording. Because technically DHCP is implemented in IPv6 and this resource will still use it when a static address is not assigned but the check to see if it is enabled is meaningless and reports that DHCP is always enabled (even if a static is assigned). This is because DHCPv6 state isn't reported in the same way as DHCPv4 - they sort of have a slightly different definition and purpose in IPv4 vs IPv6.

I hope that kind of makes sense :smile: But I'll put a warning in covering this - definitely worth doing.

PlagueHO:
When I was thinking about the appropriate warning message to show, I think there might be a better way of doing this.

I think the test for DHCP status needs to be removed completely. It shouldn't be in this resource at all.  It is as you said in Issue #17 - there should be a really simple xDHCPClient resource which simply sets DHCP property to enable/disable on the NetIPInterface (limited by AddressFamily).

The reason why I think this test is not useful here is this:
1. When a static IP Address is assigned to an IPv4 interface the DHCP Status is automatically disabled - so simply assigning an IP address sets this and there is no point in testing for it because if the assigned IP Address is different to desired IP address then Set-TargetResource will fire anyway, which will disable DHCP.

2. When a static IP Address is assigned to an IPv6 interface the DHCP Status is **not changed**.

Note: As it stands currently the xIPAddress resource has no way to enable DHCP anyway and it is never disabled by any other method except setting the IPv4 address.

So, I think I'd be much better removing this test completely and implementing an xDHCPClient resource that can be used to enable DHCP (or disable it).

Sorry about all the too-ing and fro-ing on this, but best get it right.

tysonjhayes:
Makes sense to me. Would you mind opening an issue saying we need an xDHCPServer resource specifically? I know #17 talks about it but it's a bit obscure that we actually decided to move forward on it.

PlagueHO:
Absolutely! Consider it done.

tysonjhayes:
oh bother, could you be so kind as to update the readme.md? I remembered that as soon as I clicked merge.

PlagueHO:
Done! Sorry I should have noticed that myself. Doh!

Issue #68, opened 2015-12-22
Add xDHCPClient Resource


PlagueHO:
The purpose of this resource would be to set the DHCP parameter of a specific to Enabled or Disabled.

The expected MOF:
```
[ClassVersion("1.0.0"), FriendlyName("xDHCPClient")]
class MSFT_xDHCPClient: OMI_BaseResource
{
  [Key] string InterfaceAlias;
  [Write,ValueMap{"Enabled", "Disabled"},Values{"Enabled", "Disabled"}] string State;
  [Write,ValueMap{"IPv4", "IPv6"},Values{"IPv4", "IPv6"}] string AddressFamily;
};
```

This change would address Issue #17.

At the same time this resource is released the check for DHCP Status in the xIPAddress should be removed completely as it should not be there. This is referenced by Issue #32.

I'm happy to make this change and submit a PR for this if no one has any objections?

PlagueHO:
This has been merged so I'll close it. Thanks @tysonjhayes

PR #70, opened 2015-12-22
Resolve Issue #68 - Added xDHCPClient Resource

PlagueHO:
This PR adds the xDHCPClient resource that enables DHCP to be enabled/disabled on a per Adapter/AddressFamily basis.

It also removes the test for DHCP status in the xIPAddress resource.

Unit tests are included, but Integration tests aren't because I haven't figured out a way to implement the integration test without disrupting network connectivity.

This should address Issue #32 and Issue #17.

### DIFF
--- a/DSCResources/MSFT_xDhcpClient/MSFT_xDhcpClient.psm1
+++ b/DSCResources/MSFT_xDhcpClient/MSFT_xDhcpClient.psm1
@@ -1,0 +1,173 @@
+data LocalizedData
+{
+    # culture="en-US"
+    ConvertFrom-StringData -StringData @'
+GettingDHCPClientMessage=Getting the DHCP Client on {1} interface "{0}".
+ApplyingDHCPClientMessage=Applying the DHCP Client on {1} interface "{0}".
+DHCPClientSetStateMessage=DHCP Client was set to the desired state {2} on {1} interface "{0}".
+CheckingDHCPClientMessage=Checking the DHCP Client on {1} interface "{0}".
+DHCPClientDoesNotMatchMessage=DHCP Client is not in the desired state {2} on {1} interface "{0}".
+InterfaceNotAvailableError=Interface "{0}" is not available. Please select a valid interface and try again.
+'@
+}
+
+function Get-TargetResource
+{
+    [OutputType([System.Collections.Hashtable])]
+    param
+    (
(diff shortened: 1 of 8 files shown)


DSCResources/MSFT_xDhcpClient/MSFT_xDhcpClient.psm1
```diff
@@ -0,0 +1,173 @@
...
+        ) -join '')
+
+    Test-ResourceProperty @PSBoundParameters
+    
+    $CurrentDHCPClient = Get-NetIPInterface `
+        -InterfaceAlias $InterfaceAlias `
+        -AddressFamily $AddressFamily
+        
+    # The DHCP Client is in a different state - so change it.
+    if ($CurrentDHCPClient.DHCP -ne $State)
```
tysonjhayes: Is there a scenario where this is going to be false? I ask because we only run set if test returns false.

PlagueHO: The test will return false if the DHCP state is already in the same state that the resource wants it to be. So in this case we wouldn't want to run set anyway. E.g. only fire the set if the state is different.

This is really only a double check anyway - because Test-Resource should return $true if the state is already correct - and therefore set-targetresource wouldn't even fire. 

Does this make sense? I may be missing what you mean though (not enough coffee yet).

tysonjhayes: So if Set-TargetResource shouldn't ever fire unless there is a difference why are we checking the status again? Seems redundant.

PlagueHO: It really is redundant TBH - but I put in in there "just-in-case" there was a flaw in my logic somewhere. Not that it would actually cause a problem should this set fire when it didn't need to. I'll remove it :)

PlagueHO: It has been removed.

PlagueHO: Ad fixed the unit tests :smile:

tysonjhayes:
Thanks!

PR #71, opened 2015-12-22
Fixed xDnsConnectionSuffix Get-TargetResource returned hashtable.

austoonz:
The hashtable being returned didn't match schema.mof so Get-DscConfiguration cmdlet would fail.
Updated 'DnsSuffix' to 'ConnectionSpecificSuffix'.


PlagueHO:
LGTM! :+1:

Implementing Integration tests would catch this sort of issue in future I think.

iainbrighton:
Good catch! LGTM too.

tysonjhayes:
Thanks! I've opened an issue to get integration tests written for these resources which should catch things like this in the future.

Thank you!

PR #82, opened 2016-01-19
Adding NetworkTeam resource to close #80

### LINKED ISSUE #80: [New Resource] NIC Teaming (NetLBFO)
Add [Nic Teaming (NetLBFO)](https://technet.microsoft.com/en-us/library/jj130849(v=wps.630).aspx) to the list of resources.

rchaganti:
This adheres to the style and coding guidelines of this repo and has
been updated to add examples and initial unit tests. I will add more
unit tests and integration tests very soon.
https://github.com/PowerShell/xNetworking/issues/80


DSCResources/MSFT_NetworkTeam/MSFT_NetworkTeam.psm1
```diff
@@ -0,0 +1,243 @@
...
+        teamMembers = $teamMembers
+    }
+
+    Write-Verbose ($localizedData.GetTeamInfo -f $name)
+    $networkTeam = Get-NetLBFOTeam -Name $name -ErrorAction SilentlyContinue
+
+    if ($networkTeam) 
+    {
+        Write-Verbose ($localizedData.FoundTeam -f $name)
+        if ($null -eq (Compare-Object -ReferenceObject $teamMembers -DifferenceObject $networkTeam.Members))
```
tysonjhayes: In your Mock Object $networkTeam has TeamMembers, here you call Members.  This difference is likely what's failing the test. Which one should it be?  My guess is TeamMembers but I'm not 100% sure.

DSCResources/MSFT_NetworkTeam/MSFT_NetworkTeam.psm1
```diff
@@ -0,0 +1,243 @@
...
+}
+
+Function Get-TargetResource
+{
+    [CmdletBinding()]
+    [OutputType([System.Collections.Hashtable])] 
+    Param
+    (
+        [Parameter(Mandatory)]
+        [string]$name,
```
tysonjhayes: Parameters should use [PascalCase](https://github.com/PowerShell/DscResources/blob/master/StyleGuidelines.md#parameter-names-should-use-pascalcase-) please

DSCResources/MSFT_NetworkTeam/MSFT_NetworkTeam.psm1
```diff
@@ -0,0 +1,243 @@
...
+
+        [Parameter(Mandatory)]
+        [String[]]$teamMembers
+    )
+    
+    $configuration = @{
+        name = $name
```
tysonjhayes: nit: References to Parameters should also use the same capitalization styles (PascalCase in this instance)

DSCResources/MSFT_NetworkTeam/MSFT_NetworkTeam.psm1
```diff
@@ -0,0 +1,243 @@
...
+                    Write-Verbose ($localizedData.addingMembers -f ($membersToAdd -join ','))
+                    $null = Add-NetLbfoTeamMember -Name $membersToAdd -Team $name -ErrorAction Stop -Confirm:$false
+                }
+            }
+            
+        } 
+        else 
+        {
+            Write-Verbose ($localizedData.createTeam -f $name)
+            $null = New-NetLbfoTeam -Name $name -TeamMembers $teamMembers -TeamingMode $teamingMode -LoadBalancingAlgorithm $loadBalancingAlgorithm -ErrorAction Stop -Confirm:$false
```
tysonjhayes: Could you please use some backticks (`) to break this up to multiple lines to keep it under 100 characters?

````powershell
New-NetLbfoTeam `
    -Name $Name `
   -TeamMembers $TeamMembers `
   -TeamingMode $TeamingMode `
   -LoadBalancingAlgorithm $LoadBalancingAlgorithm `
   -ErrorAction Stop `
   -Confirm:$false
````

File: DSCResources/MSFT_NetworkTeam/MSFT_NetworkTeam.psm1 [hunk 5]
```diff
@@ -0,0 +1,243 @@
...
+    
+    if ($ensure -eq 'Present')
+    {
+        if ($networkTeam)
+        {
+            Write-Verbose ($localizedData.foundTeam -f $name)
+            if (
+                ($networkTeam.LoadBalancingAlgorithm -eq $loadBalancingAlgorithm) -and 
+                ($networkTeam.teamingMode -eq $teamingMode) -and 
+                ($null -eq (Compare-Object -ReferenceObject $teamMembers -DifferenceObject $networkTeam.Members))
```
tysonjhayes: This will probably have the same problem as L58 with the difference in Members and TeamMembers

DSCResources/MSFT_NetworkTeam/MSFT_NetworkTeam.psm1
```diff
@@ -0,0 +1,243 @@
...
+    teamDoesNotExistShouldCreate=Network team with name {0} does not exist. It will be created.
+    teamExistsShouldRemove=Network team with name {0} exists. It will be removed.
+    teamDoesNotExistNoAction=Network team with name {0} does not exist. No action needed.
+    waitingForTeam=Waiting for network team status to change to up.
+    createdNetTeam=Network Team was created successfully.
+    failedToCreateTeam=Network team with specific configuration failed to changed to up state within timeout period of 120 seconds.
+'@
+}
+
+if (Test-Path "${PSScriptRoot}\${PSUICulture}")
```
tysonjhayes: Maybe here we could have a Test-Path (whatever) and an else Import en-US so we can just remove the lines above. Instead of having two different string sets (here and a seperate file).  This chunk might make a good standard to go off of too (what do you think @PlagueHO?)

Examples/Sample_NetworkTeam_AddTeam.ps1
```diff
@@ -0,0 +1,24 @@
+configuration Sample_NetworkTeam_AddTeam
```
tysonjhayes: Love that you wrote examples. Thank you! :+1:

Tests/Unit/MSFT_NetworkTeam.Tests.ps1
```diff
@@ -0,0 +1,83 @@
...
+                    $Result.Name                   | Should Be $TestTeam.Name
+                    $Result.TeamMembers            | Should Be $TestTeam.TeamMembers
+                    $Result.loadBalancingAlgorithm | Should Be 'Dynamic'
+                    $Result.teamingMode            | Should Be 'SwitchIndependent'
+                }
+                It 'should call the expected mocks' {
+                    Assert-MockCalled -commandName Get-NetLbfoTeam -Exactly 1
+                }
+            }
+        }
```
tysonjhayes: I'd really want to see tests for Get/Set resources, do you need some help with those?

tysonjhayes:
I think I found the problem with your test and made a note where I think the problem is. Logic all seems pretty sound, would really like to see some tests for the other functions.

DSCResources/MSFT_NetworkTeam/MSFT_NetworkTeam.psm1
rchaganti: My bad! I was using mock objects but was looking for the value by parameter name! :)

DSCResources/MSFT_NetworkTeam/MSFT_NetworkTeam.psm1
rchaganti: All we need here is this:
if (Test-Path "${PSScriptRoot}\${PSUICulture}")
{
    Import-LocalizedData -BindingVariable LocalizedData -filename MSFT_NetworkTeam.psd1 -BaseDirectory "${PSScriptRoot}\${PSUICulture}"
}

I can remove the lines above this.

Tests/Unit/MSFT_NetworkTeam.Tests.ps1
rchaganti: I will add these eventually! wanted to push this first and start adding more tests. Is that OK?

DSCResources/MSFT_NetworkTeam/MSFT_NetworkTeam.psm1
PlagueHO: I'm in two minds here - I like the idea of every culture being in it's own file (including en-us). But I took a look at what was done in the PSDesiredStateConfiguration Resource Module in WMF5.0 and the localization is performed like this:
```powershell
data LocalizedData
{
    # culture="en-US"
    ConvertFrom-StringData -StringData @'
    CheckSumFileExists = File '{0}' already exists. Please specify -Force parameter to overwrite existing checksum files.
    CreateChecksumFile = Create checksum file '{0}'
...
'@
}

Import-LocalizedData  -BindingVariable LocalizedData -FileName PSDesiredStateConfiguration.Resource.psd1 -ErrorAction SilentlyContinue
```
and the en-us culture is included in the base file. I think keeping the en-us in the base file is probably the way to go to just so there is at least something to easily refer back to in the core file (without having to jump back to another file to see what a particular message says).

I would say that the Import-LocalizedData needs to be split over multiple lines to come over the 100 char limit:
```powershell
Import-LocalizedData  `
   -BindingVariable LocalizedData `
   -FileName PSDesiredStateConfiguration.Resource.psd1 `
   -ErrorAction SilentlyContinue
```

DSCResources/MSFT_NetworkTeam/MSFT_NetworkTeam.psm1
```diff
@@ -0,0 +1,243 @@
...
+        [Parameter(Mandatory)]
+        [String[]]$teamMembers
+    )
+    
+    $configuration = @{
+        name = $name
+        teamMembers = $teamMembers
+    }
+
+    Write-Verbose ($localizedData.GetTeamInfo -f $name)
```
PlagueHO: Not to be a complete pain, but can you change the Write-Verbose functions to use non-positional parameters? E.g
```powershell
Write-Verbose -Message ($localizedData.GetTeamInfo -f $name)
```

A replace in your editor should be able to convert all these easily enough. :smile:

DSCResources/MSFT_NetworkTeam/en-US/MSFT_NetworkTeam.psd1
```diff
@@ -0,0 +1,22 @@
+ConvertFrom-StringData @'
```
PlagueHO: I'm not sure this file is actually needed if we keep the en-us culture in the base file. Best to avoid having to change the string in two different places?

DSCResources/MSFT_NetworkTeam/MSFT_NetworkTeam.psm1
```diff
@@ -0,0 +1,243 @@
...
+                $isNetModifyRequired = $true
+            }
+            
+            if ($isNetModifyRequired)
+            {
+                Write-Verbose ($localizedData.modifyTeam -f $name)
+                Set-NetLbfoTeam @setArguments -ErrorAction Stop -Confirm:$false
+            }
+
+            $netTeamMembers = Compare-Object -ReferenceObject $teamMembers -DifferenceObject $networkTeam.Members
```
PlagueHO: Could we split this line to come in under the 100 char limit? E.g.
```powershell
$netTeamMembers = Compare-Object -ReferenceObject $teamMembers -DifferenceObject $networkTeam.Members
```

DSCResources/MSFT_NetworkTeam/MSFT_NetworkTeam.psm1
PlagueHO: What you probably need to do here is ensure that both $teamMembers and $networkTeam.Members are actually arrays. So perform an explicit cast here to be sure:
```powershell
 if ($null -eq (Compare-Object -ReferenceObject @($teamMembers) -DifferenceObject @($networkTeam.Members)))
```

This may fix the issue.

DSCResources/MSFT_NetworkTeam/MSFT_NetworkTeam.psm1
PlagueHO: Same thing here with explicitly casting to arrays.

DSCResources/MSFT_NetworkTeam/MSFT_NetworkTeam.psm1
```diff
@@ -0,0 +1,243 @@
...
+        {
+            Write-Verbose ($localizedData.createTeam -f $name)
+            $null = New-NetLbfoTeam -Name $name -TeamMembers $teamMembers -TeamingMode $teamingMode -LoadBalancingAlgorithm $loadBalancingAlgorithm -ErrorAction Stop -Confirm:$false
+            $timeout = 0
+            While ((Get-NetLbfoTeam -Name $name).status -ne 'Up')
+            {
+                Write-Verbose $localizedData.waitingForTeam
+                if ($timeout -ge 120)
+                {
+                    throw $localizedData.failedToCreateTeam
```
PlagueHO: Instead of calling a throw here, can you use the following construct to throw the exception?
https://github.com/PowerShell/xNetworking/blob/dev/DSCResources/MSFT_xFirewall/MSFT_xFirewall.psm1#L708

I know this is a bit longer winded, but it creates a custom exception object that we can actually test for in unit tests.

So you'd end up with this:
```powershell
        $errorId = 'TeamCreateError'
        $errorCategory = [System.Management.Automation.ErrorCategory]::InvalidOperation
        $errorMessage = $localizedData.failedToCreateTeam
        $exception = New-Object -TypeName System.InvalidOperationException `
            -ArgumentList $errorMessage
        $errorRecord = New-Object -TypeName System.Management.Automation.ErrorRecord `
            -ArgumentList $exception, $errorId, $errorCategory, $null

        $PSCmdlet.ThrowTerminatingError($errorRecord)
```

PlagueHO:
This is all looking pretty awesome! :+1: Thanks @rchaganti ! Your code is really nice and easy to understand. And as @tysonjhayes said, thanks for the examples- they are great.

We will also need to update the readme.md file as well to include details of this new resource. I'm happy to help out where I can here too with documentation if required.

Tests/Unit/MSFT_NetworkTeam.Tests.ps1
tysonjhayes: Awesome! Just wasn't sure how final this was. :+1:

DSCResources/MSFT_NetworkTeam/MSFT_NetworkTeam.psm1
rchaganti: Thanks! This is a good suggestion. I added this for the complete error record.

DSCResources/MSFT_NetworkTeam/MSFT_NetworkTeam.psm1
rchaganti: I updated a few lines to go under 100 char limit. However, there are a few more than are over 100 but less than 110. I am leaving them as is for better readability.

DSCResources/MSFT_NetworkTeam/MSFT_NetworkTeam.psm1
rchaganti: Find and Replace #FTW

DSCResources/MSFT_NetworkTeam/MSFT_NetworkTeam.psm1
rchaganti: I prefer always keeping the localized data in the culture folder. I have removed inline strings but would like to retain the structure in the module folder. This is easier to have others add their localizations and not worry about anything in the PSM1 file. So, I prefer removing the inline strings and leave it only in the external culture folder.

DSCResources/MSFT_NetworkTeam/MSFT_NetworkTeam.psm1
rchaganti: Done!

TravisEz13:
There is a failing Unit test.

rchaganti:
I have not committed changes to the PR yet after all the earlier feedback. I will do so today once my local testing is complete.

rchaganti:
Finally! The tests were failing earlier because of a wrong comparison in the mock. It is now fixed. Do you want to merge this or wait for all Set/Test unit tests too! I may need a couple of more days for that.

tysonjhayes:
Personally I'd rather have all of the tests before we merge, I'm willing to wait some time for that if needed.

rchaganti:
No issues. I can get you those tests. I have some challenges with Set-TargetResource.

tysonjhayes:
Feel free to push the code and comment on where the problems are, I'm happy to help track it down with you (and I'm sure @PlagueHO would help too :grinning:). Or hit me up on the PowerShell/DSCResources chat in Gitter.  Whatever works for you.

TravisEz13:
Agreed, it is better to have the test coverage before merging.

PlagueHO:
Absolutely! I'm well keen to help out where I can! :+1:

rchaganti:
Agree. will be updating this PR as I complete all unit tests. @tysonjhayes the issue with Set-TargetResource in my resource module is that it calls Get-NetLbfoTeam initially to check if team exists and then uses the same to check if the team status is UP or not. I used that logic to wait until the team becomes functional. This has an impact on the mocks for Get-NetLbfoTeam in Set-TargetResource tests. I am doing some changes around that logic to make it more unit test friendly! Any suggestions are welcome.

rchaganti:
I added unit tests for all DSC resource functions and integration tests for the resource configuration. But, as I suspected the integration tests (for this resource especially) will break build process as creating a network team inside AppVeyor VM will break network connectivity. For the sake of integration testing, should I change ensure in the configuration from Present to Absent? It doen't make sense to me but integration tests are mandatory, this is what should happen.

README.md
```diff
@@ -111,6 +113,12 @@ Please check out common DSC Resources [contributing guidelines](https://github.c
...
 * **PreferredLifetime**: Specifies a preferred lifetime in seconds of an IP route.
 
+### NetworkTeam
+* **Name**: Specifies the name of the network team to create.
+* **TeamMembers**: Specifies the network interfaces that should be a part of the network team. This is a comma-separated list.
+* **TeamingMode**: Specifies the teaming mode configuration. { SwitchIndependent | LACP | Static}.
+* **LoadBalancingAlgorithm**: Specifies the load balancing algorithm for the network team. { Dynamic | HyperVPort | IPAddresses | MacAddresses | TransportPorts }.
+* **Ensure**: Specifies if the network team should be created or deleted. { Present | Absent }.
 
 ## Known Invalid Configurations
```
TravisEz13: Please include a summary of changes in an unreleased version change per the [contribution guidelines](https://github.com/PowerShell/DscResources/blob/master/CONTRIBUTING.md#lifecycle-of-a-pull-request)

PlagueHO:
There are some resources where Integration testing isn't possible because it would be destructive on the testing system (or would require additional servers or services which can't be reasonably/automatically provided).

In these cases I actually create the integration test script but put a ```return``` as the first line of the code so the tests don't actually fire (but they are written). This way if a user wants to run the tests on their own test machines and are OK with destructive testing then it is up to them to remove the return and deal with the consequences. The same would go for tests where additional services or servers are required.

tysonjhayes:
@rchaganti - I'm not super concerned with Integration tests for this at the moment, we need a better story for many of our resources on this as @PlagueHO is pointing out.

@PlagueHO - I thought you had some lead on creating a loopback adapter on the system, did that not pan out?

PlagueHO:
@tysonjhayes - still working on that one - got a bit side tracked with finishing the integration tests on my cDFS resources. But it's on my list :) If I could get it working then it would allow integration tests on this resource too. I'll keep you posted.

README.md
rchaganti: Added
* MSFT_NetworkTeam resource to manage native network teaming.

to unreleased list.

rchaganti:
@tysonjhayes yes, I felt the same when writing these integration tests. These are some tests that I use in a "real" environment. For CI builds, it doesn't make sense. We can probably add this note in the guidelines stating integration tests for resources like this should follow the return pattern suggested by @PlagueHO. I implemented that in my integration test script.

PlagueHO:
Nice! :+1:

TravisEz13:
Thank, I believe my issues have been addressed.

rchaganti:
@TravisEz13 you want me to rename this as well with x prefix? I have seen your comments on the Ben's PR. So, checking.

tysonjhayes:
@rchaganti - Yes, sorry about any confusion here.

TravisEz13:
I agree with Tyson

rchaganti:
Done. I renamed the resource to xNetworkTeam and updated this PR. I am moving on to HostsFile.

PlagueHO:
Nice! :+1:

DSCResources/MSFT_xNetworkTeam/MSFT_xNetworkTeam.psm1
```diff
@@ -0,0 +1,230 @@
+﻿if (Test-Path "${PSScriptRoot}\${PSUICulture}")
+{
+    Import-LocalizedData -BindingVariable LocalizedData -filename MSFT_xNetworkTeam.psd1 -BaseDirectory "${PSScriptRoot}\${PSUICulture}"
+}
```
tysonjhayes: Someone correct me here if I'm wrong but if I have a language that's not found it's not going to load any strings correct?  Shouldn't we at least have an else statement to load english?

TravisEz13: See [notes](https://technet.microsoft.com/en-us/library/hh849919.aspx)  Why don't we just let import-localizeddata do the search?

rchaganti: :) that is precisely the reason I had a dump of all scripts at the beginning of the PSM1. So, removing that completely got me into this bug. I will add the conditional load of en-US.

TravisEz13: @rchaganti 
No, I'm fine with base directory.  It is ok to give the cmdlet a hint on where to look.  But, it seem we should leave the language search up to the cmdlet (unless we have a really good reason,) as it should be well used and tested code.

FYI, you have a merge conflict

rchaganti:
I resolved the merge conflicts and committed to this PR.

tysonjhayes:
Thanks! Everything looked good to me so I merged.

@KarolKaczmarek - whenever you have time could we release this to the gallery please?

TravisEz13:
Thanks everyone!

PlagueHO:
Nice! :+1:

KarolKaczmarek:
@tysonjhayes Done!
http://www.powershellgallery.com/packages/xnetworking/2.7.0.0

PR #98, opened 2016-02-15
Adding MSFT_xHostsFile resource

rchaganti:


README.md
```diff
@@ -144,6 +144,8 @@ The cmdlet does not fully support the Inquire action for debug messages. Cmdlet
 ### Unreleased
 
 * Templates folder removed. Use the test templates in the [Tests.Template folder in the DSCResources repository](https://github.com/PowerShell/DscResources/tree/master/Tests.Template) instead.
```
TravisEz13: Need to add a description of the resources and its properties, see line 121 for xNetworkTeam
and an example line 260 after your changes starts the examples.

rchaganti: Done!

DSCResources/MSFT_xHostsFile/MSFT_xHostsFile.psm1
```diff
@@ -0,0 +1,314 @@
...
+
+    Write-Verbose -Message $localizedData.checkingHostsFileEntry
+    try
+    {
+        if (Test-HostEntry -IPAddress $IPAddress -HostName $HostName) {
+            Write-Verbose -Message ($localizedData.hostsFileEntryFound -f $HostName, $IPAddress)
+            $configuration.Add('Ensure','Present')
+        } else {
+            Write-Verbose -Message ($localizedData.hostsFileEntryNotFound -f $HostName, $IPAddress)
+            $configuration.Add('Ensure','Absent')
```
TravisEz13: Technically it can be present with a different IP address.  In getting, we should return based on the key.

TravisEz13: never mind both are the key

README.md
TravisEz13: I don't see an example added, but I don't think I'm going to block on that. Please follow up with an example.

TravisEz13:
Please follow up with an example in the readme as mentioned.

PR #105, opened 2016-03-16
Fix Test-TargetResource where False Negative reported - Fixes #104

### LINKED ISSUE #104: MSFT_xIPAddress: IP Address Test wrong result
Hi,

It seems the Test-Resource of MSFT_xIPAddress delivers a wrong result, if the desired IP address (as string) is a part of the current one.

Example: 
Current IP: 192.168.178.15
Desired IP: 192.168.178.1

The if-clause "if (-not $currentIPs.IPAddress.Contains($IPAddress))" delivers false - so DSC thinks the current IP address is equal to the desired one...

PlagueHO:
Fix Test-TargetResource where False Negative reported when desired IP is a substring of the current IP.

Unit tests updated to detect this issue.


tysonjhayes:
LGTM. Thanks!

Issue #94, opened 2016-02-05
Need a resource to set a NIC alias so you can reliably use xDnsServerAddress


TravisEz13:
xDnsServerAddress uses the NIC alias as the key.  You shouldn't have to query ahead of time the nic alias or guess

PlagueHO:
I think we could add support to use a wildcard NIC alias - e.g. '*' or 'ethernet*' . The NIC alias would still remain the key.

PlagueHO:
I think there was a PR floating around to add support for wildcard aliases? I recall there was one but there was some trouble with it so the PR got closed. @tysonjhayes, do you remember if this was the case? I can't find the Issue or PR though. Perhaps the change could be resuscitated?

iainbrighton:
Yep - there was one floating around that got closed recently. It might have been in another repo though.

I'll check the ones I'm tracking..

tysonjhayes:
Yes it was #57, was recently closed because the author didn't have time to get it fixed up to be merged.

TravisEz13:
@PlagueHO @tysonjhayes @iainbrighton You may need to query to find the NIC on multiple values (especially on physical machines, Such as the physical media type, the status, etc.  Additionally, predicting the name might be difficult if the hardware is not standardized.).  A separate resource that sets the name would be more salable (the actual interface alias could be used on every resource instead of needing to put the query code in each resource.)

I've prototyped such a resource as [xNetworkAdapterName](https://github.com/TravisEz13/xNetworking/tree/NetAdapter)

PlagueHO:
@TravisEz13, I really like your idea about the xNetworkAdapterName resource instead of needing to change xDNSServerAddress.

I'd also suggest that the xNetworkAdapterName resource should support setting the adapter name based on MAC as well. The reason for this is that I automate VM creation where I know/can set the MAC of each adapter in a VM on creation and automatically assemble the DSC configurations with that in mind. At the moment I'm using unattend.xml to configure the Adapter name on first boot so that xDNSServerAddress etc can be used against a known adapter name. So this would solve a problem for me.

My 2c. Thanks.

TravisEz13:
@PlagueHO I'm hitting design issues on the resource.  My current thought is to write the helpers to perform the work and expose them in the xNetworking module.  Then the helpers can be used in a script resource for now until the design work is done.

PlagueHO:
@TravisEz13 - this sounds like a good approach to me.

[...]

PR #110, opened 2016-04-18
Added functions to query for and rename a netadapter

TravisEz13:
This change is [Reviewable](https://reviewable.io/reviews/powershell/xnetworking/110)


TravisEz13:
Almost fixes #94

PlagueHO:
This is looking pretty good!

After reviewing it, it got me thinking in detail about exact usage patterns.

So to kick off, the parameters I think that would most likely be required are:
- MacAddress
- ifIndex
- DriverDescription

These are properties of Microsoft.Management.Infrastructure.CimInstance#ROOT/StandardCimv2/MSFT_NetAdapter

Example scenarios:
For large physical hosts, you're likely to multiple adapters of the same DriverDescription - for HA and QoS. You may have each adapter plugged into different physical networks (Management, Cluster traffic, SAN etc etc), so you'd want to name them accordingly. But they're all likely to be the same hardware - so you'd select them based on ifIndex and DriverDescription. You might select them on MACAddress, but this would require the DSC config to be customized for each Node - which you want to avoid - especially for Clusters. Physical servers may have other adapter types as well (10GbE for SAN etc) - so the DriverDescription would allow differentiating them.

All adapters in Virtual Machines will have adapters with the same DriverDescription as well - except when the host has SR-IOV adapters. So the same thing applies here - you'll want to be able to set the Adapter name based on the ifIndex and Driver.

An alternate method would be to allow a custom filter to be specified as a parameter E.g.
CustomFilter = "($_.DriverDescription -eq 'Microsoft Hyper-V Network Adapter') -and ($_.ifIndex -eq 1)"

Just a few ideas to get things going - @IainBrighton - this is something you're eager for...what are your thoughts on what you'd need?

Also, just thinking here, is it required that the other existing resources use this new Module - could we just create a new xNetworkAdapterName module using this code? Rather than allowing each existing resource to identify the adapter to work with. It seems to me that if an adapter is always forced to the correct name before any other Networking resources are applied then it might be simpler. Just throwing ideas out there. :smile:

---

Reviewed 9 of 11 files at r2.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[README.md, line 879 \[r2\]](https://reviewable.io:443/reviews/powershell/xnetworking/110#-KFhccn08-twI39u_7ZT:-KFhccn08-twI39u_7ZU:-1563386957) ([raw file](https://github.com/powershell/xnetworking/blob/e868bd0697bcfc6ca0fe18a61cbc70a636604cb5/README.md#L879)):*
Needs a new line.

---

*[xNetworkAdapter.psm1, line 17 \[r2\]](https://reviewable.io:443/reviews/powershell/xnetworking/110#-KFhcl1IDazqz2HNhc-r:-KFhcl1IDazqz2HNhc-s:-143661984) ([raw file](https://github.com/powershell/xnetworking/blob/e868bd0697bcfc6ca0fe18a61cbc70a636604cb5/xNetworkAdapter.psm1#L17)):*
Can ApplyingWhileInDesiredStateMessage be removed? It doesn't appear to be used.

---

*[xNetworkAdapter.psm1, line 138 \[r2\]](https://reviewable.io:443/reviews/powershell/xnetworking/110#-KFhd7N3V01mS7RHuTFU:-KFhd7N3V01mS7RHuTFV:1077441357) ([raw file](https://github.com/powershell/xnetworking/blob/e868bd0697bcfc6ca0fe18a61cbc70a636604cb5/xNetworkAdapter.psm1#L138)):*
Line length might be a bit long here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/xnetworking/110#-:-KFhdHUivKz-VWOeAgPg:-2080962223)*

TravisEz13:
Remembering my goal is to turn this into a resource.  I don't think we should allow custom filters.  This avoids describing the state and there is almost no way to avoid code injection in the filter without getting really complex.  Adding values you can filter on (with no default) is not a breaking change, as effectively, you can think of name as the key, even though it is the value being set (one of the design issues for the resource.)

---

Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[README.md, line 879 \[r2\]](https://reviewable.io:443/reviews/powershell/xnetworking/110#-KFhccn08-twI39u_7ZT:-KFkZAlMZYHeHF0NKOQM:1637401584) ([raw file](https://github.com/powershell/xnetworking/blob/e868bd0697bcfc6ca0fe18a61cbc70a636604cb5/README.md#L879)):*
Fixed in next commit

---

*[xNetworkAdapter.psm1, line 17 \[r2\]](https://reviewable.io:443/reviews/powershell/xnetworking/110#-KFhcl1IDazqz2HNhc-r:-KFkZKwhpq2fb32meQOA:1637401584) ([raw file](https://github.com/powershell/xnetworking/blob/e868bd0697bcfc6ca0fe18a61cbc70a636604cb5/xNetworkAdapter.psm1#L17)):*
Fixed in next commit

---

*[xNetworkAdapter.psm1, line 138 \[r2\]](https://reviewable.io:443/reviews/powershell/xnetworking/110#-KFhd7N3V01mS7RHuTFU:-KFkZxlnNzo-JFigonzn:468461402) ([raw file](https://github.com/powershell/xnetworking/blob/e868bd0697bcfc6ca0fe18a61cbc70a636604cb5/xNetworkAdapter.psm1#L138)):*
Fixed in next commit

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/xnetworking/110#-:-KFk_C58t1uaoUU1Q8JV:-564811283)*

PlagueHO:
Cool - you're quite right.

---

Reviewed 1 of 11 files at r2, 2 of 2 files at r3.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/xnetworking/110#-:-KFkiVWiGA3x43WQe9sL:-274636868)*

PR #149, opened 2016-10-03
BREAKING CHANGE - Renamed xIPAddress SubnetMask parameter to PrefixLength - Fixes #145

### LINKED ISSUE #145: SubnetMask vs. PrefixLength
I suggest to use a 'PrefixLength' parameter instead of 'SubnetMask'. A subnet mask is a 4-octect value (like 255.255.255.0) while the prefix length is an integer (from 0 to 32 for IPv4 and from 0 to 128 for IPv6). 
The current naming is confusing because the SubnetMask parameter expects a prefix length and not a subnet mask.
Getting rid of subnet masks in favor of prefix lengths is also good as subnet masks are a thing of the past in IP routing.

PlagueHO:
Changes:
* BREAKING CHANGE - MSFT_xIPAddress: SubnetMask parameter renamed to PrefixLength.

See #145

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/xnetworking/149)

### DIFF
--- a/DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1
+++ b/DSCResources/MSFT_xIPAddress/MSFT_xIPAddress.psm1
@@ -13,8 +13,8 @@ IPAddressSetStateMessage=IP Interface was set to the desired state.
 CheckingIPAddressMessage=Checking the IP Address.
 IPAddressDoesNotMatchMessage=IP Address does NOT match desired state. Expected {0}, actual {1}.
 IPAddressMatchMessage=IP Address is in desired state.
-SubnetMaskDoesNotMatchMessage=Subnet mask does NOT match desired state. Expected {0}, actual {1}.
-SubnetMaskMatchMessage=Subnet mask is in desired state.
+PrefixLengthDoesNotMatchMessage=Prefix Length does NOT match desired state. Expected {0}, actual {1}.
+PrefixLengthMatchMessage=Prefix Length is in desired state.
 DHCPIsNotDisabledMessage=DHCP is NOT disabled.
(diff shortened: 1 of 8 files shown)


lf-:
Given that this is a breaking change, should the class version not be changed to 2.0.0 instead of 1.0.0.0? Doing that makes it absolutely clear that it's incompatible.

PlagueHO:
@lf- - that did occur to me too, but I haven't been able to find specific guidance from the community on if this should be changed once it is released (even for a breaking change). For example, just adding a new property to the MOF doesn't require any version number change.

@kwirkykat - do you or your team have any thoughts on when we should update the class version of a MOF should be changed? Hopefully you haven't documented this already and I'm just being dumb and missing it.

kwirkykat:
@PlagueHO Not missing anything. There is no documentation on this since we don't touch the MOF version. If you call ```Get-DscResource``` it will just give you the version of the module. I haven't seen anywhere where the MOF version actually appears

PlagueHO:
@kwirkykat - thank you! Ok. So I'll just leave the MOF version alone. I'll fix the merge conflicts tomorrow.

PlagueHO:
@kwirkykat - fixed the merge conflict. Thank you!

kwirkykat:
:lgtm:

---

Reviewed 7 of 8 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/xnetworking/149#-:-KV6nOcJHEbfY0mj_Lxu:bnfp4nl)*

PR #345, opened 2018-07-28
IPAddress: Added new option "KeepExistingAddress"

nyanhp:
**Adds parameter KeepExistingAddress to enable scenarios where existing IP addresses should not be removed**

**This Pull Request (PR) fixes the following issues: None**

**Task list:**
- [ X ] Change details added to Unreleased section of CHANGELOG.md?
- [ X ] Added/updated documentation, comment-based help and descriptions in .schema.mof files where appropriate?
- [ X ] Examples appropriately updated?
- [ X ] New/changed code adheres to [Style Guidelines](https://github.com/PowerShell/DscResources/blob/master/StyleGuidelines.md)?
- [ X ] [Unit and (optional) Integration tests](https://github.com/PowerShell/DscResources/blob/master/TestsGuidelines.md) created/updated where possible?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+- MSFT_IPAddress:
+  - Updated to allow retaining existing addresses in order to support cluster
+    configurations as well
+
 ## 6.0.0.0
 
 - New Example 2-ConfigureSuffixSearchList.ps1 for multiple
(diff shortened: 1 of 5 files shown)


johlju:
@nyanhp Curious in which scenarios this is used. Can you explain in what scenarios the IP needs to be retained? What happens if an address is manually changed to a wrong address, will this PR correct the IP address, on the next consistency check, to the correct IP address even if `KeepExistingAddress = $true`?

johlju:
@nyanhp lets wait a bit and see what @PlagueHO's thoughts are on this.

PlagueHO:
Just have a look through and a think about this one now. I definitely understand the purpose - just want to make sure we don't end up with the resource never going into state (e.g. calling Set-TargetResource even if the resource is in state).

PlagueHO:
Wow! I dropped the ball on looking through this one. I'll try and get onto it soon.

PR #480, opened 2020-10-29
DnsServerAddress: fixes #446, simplified SET, validation error handling

phbits:
#### Pull Request (PR) description
 - Simplified Set-TargetResource by removing logic to get current value. Since wildcards are not being used, the current value doesn't need to be verified because Test-TargetResource has already done the work.
 - Added error handling to write a more descriptive error when `validation=$true` and it fails.
 - Updated Verbose message strings to provide more information to ease debugging.
 - Updated common function `Get-DnsClientServerStaticAddress` to return an empty string (`@()`) rather than `$null` if DNS Servers are not specified which fixes #446.

#### This Pull Request (PR) fixes the following issues

    - Fixes #446

#### Task list

- [ ] Added an entry under the Unreleased section of the change log in the CHANGELOG.md.
      Entry should say what was changed, and how that affects users (if applicable).
- [ ] Resource documentation added/updated in README.md in resource folder.
- [ ] Resource parameter descriptions added/updated in schema.mof
      and comment-based help.
- [ ] Comment-based help added/updated.
- [ ] Localization strings added/updated in all localization files as appropriate.
- [ ] Examples appropriately added/updated.
- [ ] Unit tests added/updated. See [DSC Resource Testing Guidelines](https://github.com/PowerShell/DscResources/blob/master/TestsGuidelines.md).
- [ ] Integration tests added/updated (where possible). See [DSC Resource Testing Guidelines](https://github.com/PowerShell/DscResources/blob/master/TestsGuidelines.md).
- [ ] New/changed code adheres to [DSC Resource Style Guidelines](https://github.com/PowerShell/DscResources/blob/master/StyleGuidelines.md) and [Best Practices](https://github.com/PowerShell/DscResources/blob/master/BestPractices.md).

---
This change is [Reviewable](https://reviewable.io/reviews/dsccommunity/networkingdsc/480)

PR #482, opened 2020-12-17
DNSServerAddress: Fixed comparison error when no DNS server is set. DSC Resource returns always false

### LINKED ISSUE #211: xDnsServerAddress does not allow clearing of primary/secondary DNS Address
Address has a ValidateNotNullOrEmpty attribute however it's a valid operation to clear it.
If using DHCP then the manual address would be set back to automatic (just like the GUI).
If using a static IP, then the manual addresses would be cleared.

rcarboneras:
#### DNSServerAddress: Fixes #211  resource does not check well DNS Server list is empty, neither clears them.

The parameter Address should do the trick 

  .PARAMETER Address
       The desired DNS Server address(es). Exclude to enable DHCP.

but in module **DSC_DnsServerAddress.psm1**, when the machine has DNS set to automatic, the following lines return an array with one $null object as member:  $currentAddress = **@($null)**

253    [String[]] $currentAddress = @(Get-DnsClientServerStaticAddress `
254       @PSBoundParameters `
255        -ErrorAction Stop)

When you don't specify an $Address parameter in the DSC resource, the $Address variable is set to an array with NO members: @(), in this lines:

243    else
244    {
245        [String[]] **$Address = @()**
246    } # if

thus, this comparison is always false ( **@()** is not equal to **@($null)** ):

258    [Boolean] **$addressDifferent** = (@(Compare-Object `
259            -ReferenceObject $currentAddress `
260            -DifferenceObject $Address `
261            -SyncWindow 0).Length -gt 0)

This is easilly fixed setting $Addess to @($null)

243    else
244    {
245        [String[]] **$Address = @($null)**
246    } # if

Tested and working in my environment :)

- Fixes #211

#### Task list
- [x] Added an entry under the Unreleased section of the change log in the CHANGELOG.md.
      Entry should say what was changed, and how that affects users (if applicable).
- [ ] Resource documentation added/updated in README.md in resource folder.
- [ ] Resource parameter descriptions added/updated in schema.mof
      and comment-based help.
- [ ] Unit tests added/updated. See [DSC Resource Testing Guidelines](https://github.com/PowerShell/DscResources/blob/master/TestsGuidelines.md).
- [ ] Integration tests added/updated (where possible). See [DSC Resource Testing Guidelines](https://github.com/PowerShell/DscResources/blob/master/TestsGuidelines.md).
- [x] New/changed code adheres to [DSC Resource Style Guidelines](https://github.com/PowerShell/DscResources/blob/master/StyleGuidelines.md) and [Best Practices](https://github.com/PowerShell/DscResources/blob/master/BestPractices.md).

---
This change is [Reviewable](https://reviewable.io/reviews/dsccommunity/networkingdsc/482)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - NetAdapterAdvancedProperty
   - Updated wiki documentation showing configuration overlap with
     NetAdapterLso - fixes [Issue #475](https://github.com/dsccommunity/NetworkingDsc/issues/475).
+- DnsServerAddress
+  - Test-TargetResource returns an incorrect value when Address parameter is not set
+    fixes [Issue #211](https://github.com/dsccommunity/NetworkingDsc/issues/211)
 
 ## [8.2.0] - 2020-10-16
 
--- a/source/DSCResources/DSC_DnsServerAddress/DSC_DnsServerAddress.psm1
+++ b/source/DSCResources/DSC_DnsServerAddress/DSC_DnsServerAddress.psm1
@@ -242,7 +242,7 @@ function Test-TargetResource
     }
     else
     {
-        [String[]] $Address = @()
+        [String[]] $Address = @($null)
     } # if
 
     # Remove the parameters we don't want to splat


PlagueHO:
Hi @rcarboneras - can you fix the merge conflicts and I'll review?

phbits:
@rcarboneras, @PlagueHO 

Using this approach produces the following:

```powershell
PS C:\> [System.String[]] $One = @()
PS C:\> [System.String[]] $Two = @($null)
PS C:\> $One.Count
0
PS C:\> $Two.Count
1
```

This will create problems since `Get-TargetResource` will use a value like `$One` whereas `Test-TargetResource` will use the `$Two` value. This will cause `Set-TargetResource` to run every time since `$One -ne $Two`.

rcarboneras:
@phbits  

> @rcarboneras, @PlagueHO
> 
> Using this approach produces the following:
> 
> ```powershell
> PS C:\> [System.String[]] $One = @()
> PS C:\> [System.String[]] $Two = @($null)
> PS C:\> $One.Count
> 0
> PS C:\> $Two.Count
> 1
> ```
> 
> This will create problems since `Get-TargetResource` will use a value like `$One` whereas `Test-TargetResource` will use the `$Two` value. This will cause `Set-TargetResource` to run every time since `$One -ne $Two`.

In my tests it works. You never call the Get-Targetresource function when testing a configuration, so Test-Targetresource returns $true and Set-Targetresource is never executed :)

phbits:
@rcarboneras - [PR #480](https://github.com/dsccommunity/NetworkingDsc/pull/480) was recently merged which I believe addresses the issue you're trying to solve. Have you tried the latest build of NetworkingDSC?

rcarboneras:
@phbits @PlagueHO 
I've tried with the latest version. Now if looks cleaner :).

Nevertheless, it has the same issue in the Test-TargetResource function than before, and it doesn'twork for me

It tries to compare
[System.String[]] $Address = @()

With 
[String[]] $currentAddress = @(Get-DnsClientServerStaticAddress `
        @PSBoundParameters `
        -ErrorAction Stop)
        
 And that is the same as compare
@() 
 with
 @($null)
 
 which returns a $false always, even when no Address parameter is passed to the function.
 
 An easy fix is changing this line 
 [System.String[]] $Address = @()

for this one

 [System.String[]] $Address = @($null)
 
I will open a new PR for this.

PR #495, opened 2021-05-18
Expose -IncludeHidden parameter

### LINKED ISSUE #499: Include hidden adapters by default?
#### Details of the scenario you tried and the problem that is occurring

We have a need to configure a setting on a hidden network adapter, easily enough handled by adding -IncludeHidden to *-NetAdapter et al. 

#### Suggested solution to the issue

Options:
1: add IncludeHidden parameter to most of the resources in NetworkingDsc, which in turn passes -IncludeHidden to the appropriate cmdlets. I started PR on this already, but then PlageHQ suggested a much simpler solution:
2. Add -IncludeHidden to all the cmdlets, anyways.

My vote is for #2 with one exception: add IncludeHidden parameter to NetAdapterName resource. All other resources key off of Name/InterfaceAlias so including hidden adapters when searching seems like it wouldn't break anything: the unique name is already specified. But, for NetAdapterName the key is Description so adding -IncludeHidden could potentially increase the search space unexpectedly if wildcards are used.

#### The DSC configuration that is used to reproduce the issue (as detailed as possible)
```powershell
# insert configuration here
```

#### Version of the DSC module that was used ('dev' if using current dev branch)

gaelicWizard:
#### Pull Request (PR) description
NetAdapter module Cmdlets have an -IncludeHidden flag which allows access to adapters which are hidden; expose that flag here.

- Fixes #499: solution 1.

#### Task list

- [x] Added an entry under the Unreleased section of the change log in the CHANGELOG.md.
      Entry should say what was changed, and how that affects users (if applicable).
- [ ] Resource documentation added/updated in README.md in resource folder.
- [ ] Resource parameter descriptions added/updated in schema.mof
      and comment-based help.
- [ ] Unit tests added/updated. See [DSC Resource Testing Guidelines](https://github.com/PowerShell/DscResources/blob/master/TestsGuidelines.md).
- [ ] Integration tests added/updated (where possible). See [DSC Resource Testing Guidelines](https://github.com/PowerShell/DscResources/blob/master/TestsGuidelines.md).
- [ ] New/changed code adheres to [DSC Resource Style Guidelines](https://github.com/PowerShell/DscResources/blob/master/StyleGuidelines.md) and [Best Practices](https://github.com/PowerShell/DscResources/blob/master/BestPractices.md).

---
This change is [Reviewable](https://reviewable.io/reviews/dsccommunity/networkingdsc/495)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Updated build to use `Sampler.GitHubTasks` - Fixes [Issue #489](https://github.com/dsccommunity/NetworkingDsc/issues/489).
 - Added support for publishing code coverage to `CodeCov.io` and
   Azure Pipelines - Fixes [Issue #491](https://github.com/dsccommunity/NetworkingDsc/issues/491).
+- Allow configuration of hidden network adapters.
 
 ## [8.2.0] - 2020-10-16
 
--- a/source/DSCResources/DSC_DefaultGatewayAddress/DSC_DefaultGatewayAddress.psm1
+++ b/source/DSCResources/DSC_DefaultGatewayAddress/DSC_DefaultGatewayAddress.psm1
@@ -292,7 +292,7 @@ function Assert-ResourceProperty
         $Address
     )
 
-    if (-not (Get-NetAdapter | Where-Object -Property Name -EQ $InterfaceAlias ))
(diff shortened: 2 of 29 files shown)


PlagueHO:
@gaelicWizard - this is quite an interesting one. Could you raise an issue for us to discuss it?

What I'd like to think about (and open for discussion) is should we just switch the `-hidden` flag on and not bother with the parameter? If you can raise an issue we can get some thoughts on this because I'm trying to find a good reason why if you know the adapter name and want to configure it in some way, why should it matter if it is hidden or not?

PR #518, opened 2022-09-03
Firewall: Fix issue #428 - DSC_Firewall

### LINKED ISSUE #428: Firewall and FirewallProfile: Add PolicyStore to allow targeting local group policy
#### Details of the scenario you tried and the problem that is occurring
Enable the "PolicyStore" parameter to be specified to permit targeting the local group policy store of the machine as well as the default persistent store when using both the Firewall and FirewallProfile resources

#### Verbose logs showing the problem
New Feature request

#### Suggested solution to the issue
Add "policystore" as a DSC property for both Firewall and FirewallProfile resources
Set the default value of the th e "policystore" parameter to be "persistentstore" to make it a non-breaking change and accommodate all existing configurations.
Add "policystore" parameter to
+ Get/Set/Test-TargetResource functions in both Resources
+ Get-FirewallRule and Get-FirewalRulleProperty helper functions in Firewall resource
+ Get/Set/New-NetFirewallRule commands in Firewall resource
+ Get/Set-NetFirewallProfile commands in FirewallProfile resource

#### The DSC configuration that is used to reproduce the issue (as detailed as possible)
New Feature request

#### The operating system the target node is running
Server 2016/2019 - New Feature request

#### Version and build of PowerShell the target node is running
New Feature request

#### Version of the DSC module that was used ('dev' if using current dev branch)
dev

I have made and tested the required changes and can happily submit a pull request with the updates to the code and localization strings.

I have also added two Pester tests to the Firewall resources for the helper function "Get-FirewallRule" when the -policystore is specified and the firewall rule is present/absent.

brajjan:
#### Pull Request (PR) description

Added `PolicyStore` parameter and read only property `PolicyStoreSourceType` - fixes [Issue #428](https://github.com/dsccommunity/NetworkingDsc/issues/428). As Get-NetFirewallRule does not return the PolicyStore property I added the PolicyStoreSourceType as a read only property in the MOF file, added it to DSC_Firewall.data.psd1 as well as a parameter for `Test-TargetResource` and `Set-TargetResource` for comparing if it is in desired state. 
- Parameter PolicyStore set to localhost returns PolicyStoreSourceType: GPO
- Parameter PolicyStore set to PersistentStore returns PolicyStoreSourceType: Local

Not updated Pester tests as my knowledge of Pester tests are close to 0

- Fixes #428

#### Task list

- [x] Added an entry to the change log under the Unreleased section of the
      file CHANGELOG.md. Entry should say what was changed and how that
      affects users (if applicable), and reference the issue being resolved
      (if applicable).
- [ ] Resource documentation added/updated in README.md.
- [x] Resource parameter descriptions added/updated in README.md, schema.mof
      and comment-based help.
- [ ] Unit tests added/updated. See [DSC Community Testing Guidelines](https://dsccommunity.org/guidelines/testing-guidelines).
- [ ] Integration tests added/updated (where possible). See [DSC Community Testing Guidelines](https://dsccommunity.org/guidelines/testing-guidelines).
- [x] New/changed code adheres to [DSC Community Style Guidelines](https://dsccommunity.org/styleguidelines).

- - -
This change is [Reviewable](https://reviewable.io/reviews/dsccommunity/networkingdsc/518)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Firewall
+  - Added `PolicyStore` parameter and read only property `PolicyStoreSourceType` - fixes [Issue #428](https://github.com/dsccommunity/NetworkingDsc/issues/428).
+
 ### Changed
 
 - Updated CHANGELOG.md
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Firewall
-  - Added `PolicyStore` parameter and read only property `PolicyStoreSourceType` - fixes [Issue #428](https://github.com/dsccommunity/NetworkingDsc/issues/428).
+  - Added `PolicyStore` parameter (defaults to PersistentStore to prevent a breaking change) and read only property `PolicyStoreSourceType` - fixes [Issue #428](https://github.com/dsccommunity/NetworkingDsc/issues/428).
 
 ### Changed
 
--- a/source/DSCResources/DSC_Firewall/DSC_Firewall.data.psd1
+++ b/source/DSCResources/DSC_Firewall/DSC_Firewall.data.psd1
@@ -1,36 +1,37 @@
 @{
     ParameterList = @(
-        @{ Name = 'Name';                Variable = 'FirewallRule';                                    Type = 'String'                  }
-        @{ Name = 'DisplayName';         Variable = 'FirewallRule';                                    Type = 'String'                  }
-        @{ Name = 'Group';               Variable = 'FirewallRule';                                    Type = 'String'                  }
-        @{ Name = 'DisplayGroup';        Variable = 'FirewallRule';                                    Type = ''                        }
-        @{ Name = 'Enabled';             Variable = 'FirewallRule';                                    Type = 'String'                  }
-        @{ Name = 'Action';              Variable = 'FirewallRule';                                    Type = 'String'                  }
-        @{ Name = 'Profile';             Variable = 'FirewallRule';                                    Type = 'Array'; Delimiter = ', ' }
-        @{ Name = 'Direction';           Variable = 'FirewallRule';                                    Type = 'String'                  }
-        @{ Name = 'Description';         Variable = 'FirewallRule';                                    Type = 'String'                  }
-        @{ Name = 'RemotePort';          Variable = 'properties';   Property = 'PortFilters';          Type = 'Array'                   }
-        @{ Name = 'LocalPort';           Variable = 'properties';   Property = 'PortFilters';          Type = 'Array'                   }
-        @{ Name = 'Protocol';            Variable = 'properties';   Property = 'PortFilters';          Type = 'String'                  }
-        @{ Name = 'Program';             Variable = 'properties';   Property = 'ApplicationFilters';   Type = 'String'                  }
-        @{ Name = 'Service';             Variable = 'properties';   Property = 'ServiceFilters';       Type = 'String'                  }
-        @{ Name = 'Authentication';      Variable = 'properties';   Property = 'SecurityFilters';      Type = 'String'                  }
-        @{ Name = 'Encryption';          Variable = 'properties';   Property = 'SecurityFilters';      Type = 'String'                  }
-        @{ Name = 'InterfaceAlias';      Variable = 'properties';   Property = 'InterfaceFilters';     Type = 'Array'                   }
-        @{ Name = 'InterfaceType';       Variable = 'properties';   Property = 'InterfaceTypeFilters'; Type = 'String'                  }
-        @{ Name = 'LocalAddress';        Variable = 'properties';   Property = 'AddressFilters';       Type = 'ArrayIP'                 }
-        @{ Name = 'LocalUser';           Variable = 'properties';   Property = 'SecurityFilters';      Type = 'String'                  }
-        @{ Name = 'Package';             Variable = 'properties';   Property = 'ApplicationFilters';   Type = 'String'                  }
-        @{ Name = 'Platform';            Variable = 'FirewallRule';                                    Type = 'Array'                   }
-        @{ Name = 'RemoteAddress';       Variable = 'properties';   Property = 'AddressFilters';       Type = 'ArrayIP'                 }
-        @{ Name = 'RemoteMachine';       Variable = 'properties';   Property = 'SecurityFilters';      Type = 'String'                  }
-        @{ Name = 'RemoteUser';          Variable = 'properties';   Property = 'SecurityFilters';      Type = 'String'                  }
-        @{ Name = 'DynamicTransport';    Variable = 'properties';   Property = 'PortFilters';          Type = 'String'                  }
-        @{ Name = 'EdgeTraversalPolicy'; Variable = 'FirewallRule';                                    Type = 'String'                  }
-        @{ Name = 'IcmpType';            Variable = 'properties';   Property = 'PortFilters';          Type = 'Array'                   }
-        @{ Name = 'LocalOnlyMapping';    Variable = 'FirewallRule';                                    Type = 'Boolean'                 }
-        @{ Name = 'LooseSourceMapping';  Variable = 'FirewallRule';                                    Type = 'Boolean'                 }
-        @{ Name = 'OverrideBlockRules';  Variable = 'properties';   Property = 'SecurityFilters';      Type = 'Boolean'                 }
-        @{ Name = 'Owner';               Variable = 'FirewallRule';                                    Type = 'String'                  }
+        @{ Name = 'Name';                  Variable = 'FirewallRule';                                    Type = 'String'                  }
+        @{ Name = 'DisplayName';           Variable = 'FirewallRule';                                    Type = 'String'                  }
+        @{ Name = 'Group';                 Variable = 'FirewallRule';                                    Type = 'String'                  }
+        @{ Name = 'DisplayGroup';          Variable = 'FirewallRule';                                    Type = ''                        }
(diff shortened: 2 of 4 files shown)


johlju:
@brajjan when you are ready, take this PR out of draft so it can be reviewed.

brajjan:
Not sure if the PolicyStoreSourceType property needs to be there. My initial thought was to include that one to compare with the value of the PolicyStore parameter as the PolicyStore property does not get returned when running Get-TargetResource (or Get-NetFirewallRule). 

The updated code works as intended. If you do not set the PolicyStore parameter it will default to 'PersistentStore' (= local rule, default before adding the PolicyStore property - so no breaking change)

If you set the PolicyStore to 'localhost' it will create a local GPO rule on the active node. Verify with looking at rsop or gpedit or looking at the rule in wf.msc. Get, Set and Test all works with PersistentStore or localhost

Info about the different PolicyStores can be found here
https://learn.microsoft.com/en-us/powershell/module/netsecurity/new-netfirewallrule?view=windowsserver2022-ps#-policystore

As the property does not get returned from Get-TargetResource I am not sure how to do the proper unit tests. Also my pester skills are not great but if you could have a look it would be great. 
```
$invokeParams = @{
  Name   = 'Firewall'
  Module = 'NetworkingDsc'
  Method = 'GET'
  Property = @{
    Name          = '_GPOTEST'
    PolicyStore   = 'localhost'
    LocalPort     = @('1234')
    RemoteAddress = @('1.1.1.1')
    Protocol      = 'TCP'
  }
}

Invoke-DscResource @invokeParams                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  ConfigurationName     : 
ModuleName            : NetworkingDsc
ModuleVersion         : 0.0.1
PsDscRunAsCredential  :
ResourceId            :
SourceInfo            :
Action                : Allow
Authentication        : NotRequired
Description           :
Direction             : Inbound
DisplayGroup          :
DisplayName           : _GPOTEST
DynamicTransport      : Any
EdgeTraversalPolicy   : Block
Enabled               : True
Encryption            : NotRequired
Ensure                : Present
Group                 :
IcmpType              : {Any}
InterfaceAlias        : {Any}
InterfaceType         : Any
LocalAddress          : {Any}
LocalOnlyMapping      : False
LocalPort             : {1234}
LocalUser             : Any
LooseSourceMapping    : False
Name                  : _GPOTEST
OverrideBlockRules    : False
Owner                 :
Package               :
Platform              : {}
PolicyStore           :
PolicyStoreSourceType : GroupPolicy
Profile               : {Any}
Program               : Any
Protocol              : TCP
RemoteAddress         : {1.1.1.1}
RemoteMachine         : Any
RemotePort            : {Any}
RemoteUser            : Any
Service               : Any
PSComputerName        : localhost
```

johlju:
In Get we must evaluate the correct value for `$PolicyStore`. If the current state is that it is a GPO rule then that should return  the value `localhost`. But should we hardcode the value to 'localhost'`?  According to the link it can be set to a "path" as well? Maybe it should be a separate property for the "path" that is used when PolicyStore is set to `ActiveStore`? I see that `Get-NetFirewallRule` takes the parameter `PolicyStoreSourceType`. Can that be used to determain what store it is and return the correct value.

brajjan:
Yes, the [PolicyStoreSourceType ](https://learn.microsoft.com/en-us/powershell/module/netsecurity/get-netfirewallrule?view=windowsserver2022-ps#-policystoresourcetype) will return `GroupPolicy` if set to localhost and `Local` if set to PersistentStore. So that could be used to check the correct PolicyStore.

The [PolicyStore](https://learn.microsoft.com/en-us/powershell/module/netsecurity/new-netfirewallrule?view=windowsserver2022-ps#-policystore) parameter can be set to target central GPOs as well. I have not included that but it could be done I guess. One problem that could occur is that the `New-NetFirewallRule` command does not include sending credentials. Will probably be a problem as the credentials used for applying the code probably will not be the same as an account with GPO-write permissions in AD. Could target a few more stores. See test below - they will all list what `PolicyStoreSourceType` will return

But targeting the PersistentStore and Local GPO will always work in Get/Set/Test on Windows - and targeting the GPO store (local or central does not matter) is the main objective here as many AD-joined machines does not allow local rules to be applied.

```
foreach ($obj in $stores) { New-NetFirewallRule -Name "_Test-$obj" -DisplayName "_Test-$obj" -PolicyStore $obj -Enabled False -Direction Inbound -LocalPort '1234' -RemoteAddress '1.1.1.1' -Protocol TCP -Verbose }
VERBOSE: New-NetFirewallRule DisplayName: _Test-PersistentStore

Name                          : _Test-PersistentStore
DisplayName                   : _Test-PersistentStore
Description                   :
DisplayGroup                  :
Group                         :
Enabled                       : False
Profile                       : Any
Platform                      : {}
Direction                     : Inbound
Action                        : Allow
EdgeTraversalPolicy           : Block
LooseSourceMapping            : False
LocalOnlyMapping              : False
Owner                         :
PrimaryStatus                 : OK
Status                        : The rule was parsed successfully from the store. (65536)
EnforcementStatus             : NotApplicable
PolicyStoreSource             : PersistentStore
PolicyStoreSourceType         : Local
RemoteDynamicKeywordAddresses :

VERBOSE: New-NetFirewallRule DisplayName: _Test-ActiveStore
Name                          : _Test-ActiveStore
DisplayName                   : _Test-ActiveStore
Description                   :
DisplayGroup                  :
Group                         :
Enabled                       : False
Profile                       : Any
Platform                      : {}
Direction                     : Inbound
Action                        : Allow
EdgeTraversalPolicy           : Block
LooseSourceMapping            : False
LocalOnlyMapping              : False
Owner                         :
PrimaryStatus                 : OK
Status                        : The rule was parsed successfully from the store. (65536)
EnforcementStatus             : NotApplicable
PolicyStoreSource             : ActiveStore
PolicyStoreSourceType         : Dynamic
RemoteDynamicKeywordAddresses :

VERBOSE: New-NetFirewallRule DisplayName: _Test-localhost
Name                          : _Test-localhost
DisplayName                   : _Test-localhost
Description                   :
DisplayGroup                  :
Group                         :
Enabled                       : False
Profile                       : Any
Platform                      : {}
Direction                     : Inbound
Action                        : Allow
EdgeTraversalPolicy           : Block
LooseSourceMapping            : False
LocalOnlyMapping              : False
Owner                         :
PrimaryStatus                 : OK
Status                        : The rule was parsed successfully from the store. (65536)
EnforcementStatus             : NotApplicable
PolicyStoreSource             :
PolicyStoreSourceType         : GroupPolicy
RemoteDynamicKeywordAddresses :

VERBOSE: New-NetFirewallRule DisplayName: _Test-RSOP
New-NetFirewallRule: Access is denied.
VERBOSE: New-NetFirewallRule DisplayName: _Test-SystemDefaults
New-NetFirewallRule: Access is denied.
VERBOSE: New-NetFirewallRule DisplayName: _Test-StaticServiceStore
New-NetFirewallRule: Access is denied.
VERBOSE: New-NetFirewallRule DisplayName: _Test-ConfigurableServiceStore
Name                          : _Test-ConfigurableServiceStore
DisplayName                   : _Test-ConfigurableServiceStore
Description                   :
DisplayGroup                  :
Group                         :
Enabled                       : False
Profile                       : Any
Platform                      : {}
Direction                     : Inbound
Action                        : Allow
EdgeTraversalPolicy           : Block
LooseSourceMapping            : False
LocalOnlyMapping              : False
Owner                         :
PrimaryStatus                 : OK
Status                        : The rule was parsed successfully from the store. (65536)
EnforcementStatus             : NotApplicable
PolicyStoreSource             : No Policy Store (Invalid)
PolicyStoreSourceType         : None
RemoteDynamicKeywordAddresses :
```

PlagueHO:
Cross posted from Slack:
Hi @brajjan, Thank you for submitting the PR's! I apologize it's taking me so long to get to these (unfortunately completely slammed in my day job).

My thinking is that there are two parameters here that can be treated separately: `PolicyStore` and `PolicyStoreSourceType`.

# Policy Store
From my reading of the docs, this parameter could take on a lot of different values (as I think @johlju points out): https://learn.microsoft.com/en-us/powershell/module/netsecurity/get-netfirewallrule?view=windowsserver2019-ps#-policystore

However, this potentially should be a key value (which would be a breaking change because can't have a default value for a key). Technically the same rule could need to be created in different policy stores. It's an edge case, but if we're adding the parameter then people may try to do this - which would result in invalid config (unless you created the rule with a different name for each policy).

Either way, we'd want to add some specific tests to cover this because I don't think the data driven tests will cut it. An integration test to verify usage of different policy stores (if possible - not sure though).

# PolicyStoreSourceType
This just seems like a simple read only property - so not a big deal to add.

Just my initial thoughts on this though - haven't dug in too deep on this.

johlju:
> Technically the same rule could need to be created in different policy stores. 

If so then I agree it have to be a key. 

Just athought. To prevent making a breaking change, could we do new (preferably class) resources instead, having one resource for each policy store type? Then we could extend to more policy store types by adding more resources.

PlagueHO:
That's an interesting approach @johlju. Didn't consider this case as a side-benefit of implementing class-based resources. I'd definitely like to see this happen.

However, given the scope of a change to class-based (for just the DSC_Firewall resource), I'd suggest that it's split into two PRs:
1. Convert DSC_Firewall to class-based.
2. Add the new Policy version.

From memory, the DSC_Firewall is reasonably well abstracted already, so much of the code is not in the Get, Test, Set functions, which should make it easier to convert.

johlju:
I agree with that approach, that it is best to convert the existing resources as-is and re-use the existing tests also as-is (though they will need to change somewhat because how the methods are called) to verify that no  functionality was broken.

PlagueHO:
@brajjan - does what we're talking about make sense? The proposal is to convert the Firewall resource to class-based and then enable the creation of a new class-based resources for each Firewall Policy source (presumably only one for now). These class-based resources would share 99% code. This would eliminate the breaking change and would improve the overall resource structure.

It will mean that the conversion of Firewall to class-based would need to happen first. **Would happily take a PR for this as** I'm not likely to get to this any time soon due to too many other commitments.

brajjan:
Yeah sure, that would be great. I'll try to get some time of to look at it closer and convert it to a class based resource. Might take some time though

